PR #23, opened 2018-03-10
Contributor submission test and view refactor

cheeseweasel:
This PR adds a functional test for the happy path with contributor submissions, it also tidies up the `_form` partial.

I ran into weird behaviour when I was writing the test - data was disappearing from the database mid test, it seems transactions were not behaving as they should. To test whether it was version dependent, I tried upgrading Rails to 5.1.5, and the problem went away. That is why there is also an upgrade of Rails in there, which I realise is no small thing.

Anyway let me know your thoughts.


hanchang:
Def a bit scary to see all the gem updates but I played around with it locally and all seems to be well.

@MythicalFish Heads up that there's some minor CSS changes in here since I realize you're working on the refactor there, hope it doesn't affect you too much.

hanchang:
@cheeseweasel Couple of questions:

1. Is there a way to make this run via `rails test`? Right now it only seems to work if I specify the actual test file (i.e. `rails test test/system/contributor_submissions_test.rb`)
2. Is there a way to make this run completely headlessly? Right now when I run the test it spins up a Chrome window.

cheeseweasel:
@hanchang as mentioned yesterday, we can run the system tests with `rails test:system`. Is it worth getting something like circleci running, or Heroku CI though I'm not sure about the status of that at the moment?

As for running tests headlessly, yes sorry I forgot to switch that back on, I have done now!

hanchang:
@cheeseweasel Yes please, let's definitely incorporate Heroku CI:
https://devcenter.heroku.com/articles/heroku-ci

Thanks for switching back to headless! However, when I run the test locally via `rails test:system` I get the following error:
```
Error:
ContributorSubmissionsTest#test_can_submit_content:
ActionController::RoutingError: No route matches [GET] "/undefined/undefined"
```

Not sure what's going on here, perhaps the recent merge may have affected this?

Also, we may need to wait until https://github.com/coinfi/coinfi/pull/25 has been merged to ensure that the autodeploy works properly - @MythicalFish can speak more to the status here.

PR #27, opened 2018-03-25
Coinfi prices api

cheeseweasel:
This PR includes changes to use a separate API for price history.

The url host for the API endpoint it set in an env variable - `COINFI_PRICES_URL`, this has been added to the `.env.example` file.

To run this locally you will need the https://github.com/coinfi/coinfi-prices app running locally, this is set up to run on port 4000 in development so as not to conflict with the platform app on 3000.

Some price history will need to be ingested in the price history app for this to work, an example to get this working for the `neo` coin is as follows:

```
rails ingest:coins
rails c
> neo = Coin.find_by_symbol('NEO')
> CryptoCompare::PriceHistoryService.new.load_history_per_currency(new, :daily)
```
This should load daily history for NEO, and visiting `http://localhost:3000/coins/neo` should then display the price history chart.


app/views/layouts/application.html.erb
```diff
@@ -15,6 +15,9 @@
     <link rel="manifest" href="/manifest.json">
     <link rel="apple-touch-icon" href="<%= image_path('apple-touch-icon.png') %>">
     <link rel="icon" href="<%= image_path('favicon.png') %>">
+    <%= javascript_tag do %>
+      window.pricesURL = '<%= ENV.fetch('COINFI_PRICES_URL', '') %>';
+    <% end %>
```
cheeseweasel: @hanchang I am really not sure this is the best way to achieve this, and am open to other suggestions. I was thinking data attributes possibly, but this could be used in a number of places potentially.

However, if we're moving to react, I'm not sure it will be an issue for long.

hanchang: Yup that works for me, thanks for checking!

PR #28, opened 2018-03-26
ICO Data setup

cheeseweasel:
This PR is intended to update the data structure, and implement any other changes required for the ICO data integration.

### DIFF
--- a/db/migrate/20180326043054_add_ico_fields_to_coins.rb
+++ b/db/migrate/20180326043054_add_ico_fields_to_coins.rb
@@ -1,0 +1,70 @@
+class AddIcoFieldsToCoins < ActiveRecord::Migration[5.1]
+  def change
+    add_column :coins, :ico_status, :string
+
+    add_column :coins, :ico_usd_raised, :bigint
+    add_column :coins, :ico_start_date, :bigint
+    add_column :coins, :ico_end_date, :bigint
+
+    add_column :coins, :ico_token_price_usd, :decimal, precision: 10, scale: 2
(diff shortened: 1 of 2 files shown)


db/migrate/20180326043054_add_ico_fields_to_coins.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    add_column :coins, :ico_tokens_sold, :decimal, precision: 32, scale: 16
+
+    add_column :coins, :ico_returns_usd, :float
+    add_column :coins, :ico_returns_btc, :float
+    add_column :coins, :ico_returns_eth, :float
+
+    add_column :coins, :influencer, :jsonb
+
+    add_column :coins, :excluded_countries, :jsonb, array: true
```
hanchang: @cheeseweasel I'm actually not sure what the `array: true` option does? My Google-fu is failing me as well, mind enlightening?

cheeseweasel: It should set it up as an array of jsonb, did I misinterpret your intent here?

hanchang: Ah, I see, I think that should be fine. At the moment I don't think we'll use objects in the array, but I guess this approach makes it more future-proof in that we could introduce objects:

```
excluded_countries: [{
  symbol: 'CN',
  name: 'China',
  notes: 'Lorem ipsum'
}, {
  symbol: 'US',
  name: 'United States',
  notes: 'Dolor sit amet'
}]
```

PR #35, opened 2018-04-01
[WIP] Watchlist page

MythicalFish:


Gemfile.lock
```diff
@@ -441,6 +443,7 @@ DEPENDENCIES
   devise
   dotenv-rails
   email_address
+  fast_jsonapi
```
hanchang: Did this creep back in even after removing the `example_serializer`?

app/controllers/api/concerns/responses.rb
```diff
@@ -0,0 +1,37 @@
+module Api::Concerns::Responses
```
hanchang: Nice, super clean ✨

app/controllers/api/watchlist/concerns.rb
```diff
@@ -0,0 +1,17 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :set_watchlist
+  end
+
+  private
+
+  def set_watchlist
+    @watchlist = current_user.watchlist
```
hanchang: Maybe just

    @watchlist = current_user.watchlist || Watchlist.create(user: current_user)

app/javascript/App/containers/WatchlistContainer/normalize.js
```diff
@@ -0,0 +1,12 @@
+import { normalize, schema } from 'normalizr'
```
hanchang: This looks cool, it's new to me, what does a normalizer do in JS?

app/views/layouts/application.html.erb
```diff
@@ -16,6 +16,7 @@
     <link rel="apple-touch-icon" href="<%= image_path('apple-touch-icon.png') %>">
     <link rel="icon" href="<%= image_path('favicon.png') %>">
     <%= stylesheet_link_tag 'application' %>
+    <%= csrf_meta_tags %>
```
hanchang: Wow, how was this missing?

Gemfile.lock
MythicalFish: Gone now :)

app/controllers/api/watchlist/concerns.rb
MythicalFish: Ah yeah , much cleaner :)

app/javascript/App/containers/WatchlistContainer/normalize.js
MythicalFish: First time I'm using normalizr, but basically it's bad practice to nest objects in JSON (assuming you ever want to manipulate it), so what this does is move all nested objects to the top level with the object IDs as keys, and adds references to these keys where need be.

app/views/layouts/application.html.erb
MythicalFish: I only noticed because I needed it for the local API :D
I don't think it's strictly necessary except for certain things like AJAX; your forms already had authenticity_tokens

PR #40, opened 2018-04-03
Add ICO max and available supply fields

cheeseweasel:
This PR adds the max and available supply fields for ICO data


hanchang:
Yikes, these fields exist already as `max_supply` and `available_supply`.

PR #44, opened 2018-04-06
[PLA-45] Set ico status for listed coins

cheeseweasel:
Also don't load market data from Coin Market Cap unless a coin is listed


MythicalFish:
Looks good to me

PR #47, opened 2018-04-06
[PLA-72] Add watch button to coins index

MythicalFish:


app/helpers/watchlist_helper.rb
```diff
@@ -0,0 +1,9 @@
+module WatchlistHelper
+
+  def user_is_watching? coin
+    # Return truthy string when true (string needed for passing as data attr)
```
hanchang: Love it, thanks for the explanation!

hanchang:
💯

PR #54, opened 2018-04-08
[PLA-86] Add daily update task for supply data

cheeseweasel:


hanchang:
@cheeseweasel We should probably use the Northpole proxy instead for this.

cheeseweasel:
@hanchang Sorry I missed your previous update here! The reason it is not using the Northpole proxy is that the northpole proxy does not include max supply.

This is not something that needs to be run often however (daily would be plenty), so we can possibly get away with going direct?

PR #62, opened 2018-04-13
[PLA-118] Update previous name when name changes

cheeseweasel:


hanchang:
LGTM, thanks for the test!

app/models/coin.rb
```diff
@@ -25,6 +27,10 @@ class Coin < ApplicationRecord
     end
   end
 
+  def update_previous_name
+    self.previous_name = name_was if name_changed?
```
hanchang: Cool, I didn't know about `field_was`, nice one!

PR #65, opened 2018-04-16
PLA-133 Ingest of RSS News Feeds

hanchang:


db/schema.rb
```diff
@@ -239,6 +252,36 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "news_item_raws", force: :cascade do |t|
```
hanchang: Perhaps may be worthwhile to rename to `raw_news_items`?

PR #67, opened 2018-04-20
[PLA-154] Don't query market data API without a slug

cheeseweasel:


hanchang:
💯 Nice one

hanchang:
Theoretically we shouldn't have a `Coin` with `slug` that's an empty string though?

cheeseweasel:
Theoretically we shouldn't, but we did (due to the fact there was no unique
index, and the platform app doesn't run the latest FriendlyId which doesn't
allow blank slugs). I figured it doesn't hurt to just be explicit about the
slug here.

PR #69, opened 2018-04-26
PLA-146 Changed underscores to dashes in paths and removed some cruft routes

gduquesnay:

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -26,8 +26,8 @@ Rails.application.routes.draw do
     end
     resources :articles
     resources :users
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -26,8 +26,8 @@
     end
     resources :articles
     resources :users
-    resources :submission_categories
-    resources :contributor_submissions
+    resources :submission_categories, :path => '/submission-categories'
```
hanchang: Let's use the colon notation instead of stabby as much as possible, thanks!

PR #73, opened 2018-04-30
PLA-175 Alerting and Monitoring for FeedSources

gduquesnay:


app/controllers/webhooks/websubs_controller.rb
```diff
@@ -3,15 +3,16 @@ class Webhooks::WebsubsController < ApplicationController
...
   def superfeedr_ingest
     items = params[:items]
-    head :ok and return unless items.present?
+    head :ok and return if items.blank?
 
     items.each do |item|
       NewsItemRaw.ingest!(item, params[:source])
     end
 
+    puts "Received #{items.count} NewsItems from SuperFeedr."
```
hanchang: @gduquesnay I just realized, is Papertrail dependent on this string to alert us of the situation in case there's been no updates from SuperFeedr?

If so, could you please change the string that Papertrail is looking for (on all environments, i.e. staging and production) and also add a comment that explains what dependencies this string has? Thanks!

gduquesnay: Yep.  That was the plan.  While waiting on this to get to production though i had set up the papertrail alert based on the action name "superfeedr_ingest" which has actually turned out to be unique enough.  So Im considering leaving this line as is (with your update) just for informational purposes to easily see how many items are in the request.

hanchang: Nice nice, that's reassuring to know that there's no linkage between this line and Papertrail, good stuff.

hanchang:
LGTM!

PR #74, opened 2018-05-01
[PLA-180] Rename prod2dev to staging2dev and modify DB import source to staging…

jacobi-78:
Rename prod2dev to staging2dev and modify DB import source to staging Heroku app


hanchang:
Great stuff, LGTM!

PR #79, opened 2018-05-04
PLA-85 Update roadmap

gduquesnay:
Signed-off-by: Daniel Oduonye <contactdanny19@gmail.com>

### DIFF
--- a/app/views/pages/about.html.slim
+++ b/app/views/pages/about.html.slim
@@ -118,12 +118,8 @@ section.phr.pv4.pv6-m.bg-bright-gray.white-90
   .container
     .row.pv5.items-center
       .col-xs-12.col-lg-8.col-lg-offset-2
-        h1.white
-          | Roadmap
-        p.text-muted.smaller1
-          | The roadmap outlined below is subject to change based on user feedback, market conditions, and the competitive
-          | landscape; it is merely intended as an overview to explain the features that CoinFi intends to implement on
-          | the platform.
+        h1.white Roadmap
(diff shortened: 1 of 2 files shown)


app/views/pages/about.html.slim
```diff
@@ -114,3 +114,84 @@ section#team.phr.bg-white.tiber
...
         p Holds a Juris Doctor from UAlberta Law. 7+ years of experience as a growth hacker in media startups as well as developing multiple web properties that have cumulatively grown from zero to 10M+ pageviews. Co-founder of a keyword intelligence platform for digital marketers, acquired in 2016.
         = render partial: 'linkedin_icon', locals: {url: 'https://www.linkedin.com/in/nate-tsang-3a493a153/'}
+section.phr.pv4.pv6-m.bg-bright-gray.white-90
+  .container
+    .row.pv5.items-center
+      .col-xs-12.col-lg-8.col-lg-offset-2
+        h1.white
+          | Roadmap
+        p.text-muted.smaller1
+          | The roadmap outlined below is subject to change based on user feedback, market conditions, and the competitive
```
hanchang: @gduquesnay Hm, I'm getting whitespace issues on the review app; might be better to keep this all on one unbroken line.

gduquesnay: Wow.  Good catch.  How did you even notice that.

Will change it to single line.  

I personally prefer to break up these long lines like this (see snippet bellow).  Looking at the rest of the file though it seems like we've standardised on using single lines so I'll follow the convention for now :).

```
          ' The roadmap outlined below is subject to change based on user feedback, market conditions, and the competitive
          ' landscape; it is merely intended as an overview to explain the features that CoinFi intends to implement on
          ' the platform.
```

hanchang:
🏆

PR #81, opened 2018-05-04
[PLA-135] Ensure canonical urls

cheeseweasel:
This PR is intended to address the fact that `/coins/` & `/coins` (and indeed any other urls with a trailing slash) both render the same content. We should ensure that we have one canonical url, and non canonical urls redirect to canonical.


hanchang:
❇️

PR #92, opened 2018-05-09
Fix window.pricesURL

MythicalFish:


MythicalFish:
@hanchang Not sure why I didn't spot this before, but the `pricesURL` var was missing from fetchData because I moved the JS to the head.

PR #96, opened 2018-05-17
PLA-211 Move Google Analytics to <head>

gduquesnay:
https://app.asana.com/0/683768070860536/687760835972516


hanchang:
Hi @gduquesnay, the build failure is a result of moving `/config/webpacker.yml` to `/config/webpacker.yml.sample` - just curious for the reason behind the change?

gduquesnay:
@hanchang webpacker.yml configures the webpack dev server to use localhost which doesn't work on pairing (unless we use ssh tunneling)  I need to update my local (i.e. on pairing server) webpacker.yml but didn't want to break everyone else's config.

PR #99, opened 2018-05-24
Pla 223/import articles into newsitems

gduquesnay:


lib/tasks/data_migrations/02_import_articles_into_news_items.rake
```diff
@@ -0,0 +1,24 @@
...
+      NewsItem.create!(
+        feed_source: old_article_feed_source,
+        feed_item_id: article.url,
+        url: article.url,
+        title: article.title,
+        summary: article.summary,
+        feed_item_published_at: article.published_date,
+        feed_item_updated_at: article.updated_at,
+      )
+    end
```
hanchang: @gduquesnay Do we also need to add a line to set the `is_active` field to true for all `FeedSources`?

hanchang: We may also want to note in a comment that this data migration should be run after running `20180524092136`.

lib/tasks/feed_management.rake
```diff
@@ -36,20 +36,29 @@ namespace :feeds do
...
+  #FIXME This isn't a reliable way to parse the twitter_user because some of them
+  #have twitter urls like "https://twitter.com/cardanostiftung?lang=en"
+  #not going to fix it now because not sure if we'll ever use this again
   desc "Build Feeds Source for all coins twitter feeds"
   task :build_all_feed_sources_for_twitter  => :environment do
     #using batch_process here just for error_handling e.g. since we have manually created one of these already
     #we just want it to skip the unique constraint error that will be thrown and keep going to create the rest
     batch_process(Coin.where.not(twitter: nil).where.not(twitter: "")) do |coin|
-      FeedSource.create_from_coins_twitter!(coin)
+    twitter_user = coin.twitter.split("/").last
```
hanchang: Needs tabbing for proper alignment please. 👾

hanchang: Just to note, this was caught by CodeClimate, so please do pay attention to the build!

lib/tasks/data_migrations/02_import_articles_into_news_items.rake
gduquesnay: @hanchang  I usually have rails db:migrations automated but these data migration have to be run manually post deploy so I can usually depend on the db:migration(s) they depend on to have already been run.  Is that not the case here?

lib/tasks/feed_management.rake
gduquesnay: @hanchang sorry .... not seeing what you were referring to here.

gduquesnay: @hanchang Ah... I see it now.  Thanks.

db/schema.rb
```diff
@@ -410,4 +418,6 @@
   add_foreign_key "feed_sources", "coins"
   add_foreign_key "influencer_reviews", "coins", on_delete: :cascade
   add_foreign_key "influencer_reviews", "influencers", on_delete: :cascade
+  add_foreign_key "mentions", "coins"
+  add_foreign_key "mentions", "news_items"
```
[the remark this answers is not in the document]
gduquesnay: Right.  Sorry.  Oops.

MythicalFish:
Deleting, as this was merged into https://github.com/coinfi/coinfi/pull/97

MythicalFish:
Deleting, as this was merged into #97

PR #100, opened 2018-05-24
keep imports consistent across the application

cryptoB0T:


cryptoB0T:
@hanchang in some places we see `React.Component` while in other places we see `Component`. 

this PR makes this consistent.

hanchang:
LGTM! @MythicalFish thoughts?

MythicalFish:
Yep looks good

PR #103, opened 2018-05-28
redirect icos to icos/upcoming #687754721014226

cryptoB0T:
/icos throws 500 error
https://app.asana.com/0/683768070860536/687754721014226/f

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -48,6 +48,8 @@ Rails.application.routes.draw do
   end
 
   resources :coins, only: [:index, :show]
+  get '/icos', to: redirect('/icos/upcoming')
+  resources :coins, only: [:index, :show]
   get '/icos(/:status)', to: 'icos#index'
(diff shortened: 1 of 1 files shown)


config/routes.rb
```diff
@@ -47,6 +47,8 @@
     get '/social_feeds/tweets_by_user', to: 'social_feeds#tweets_by_user'
   end
 
+  resources :coins, only: [:index, :show]
```
hanchang: Er @cryptoB0T I think this is a duplicate line?

config/routes.rb
```diff
@@ -47,6 +47,8 @@
...
+  get '/icos', to: redirect('/icos/upcoming')
```
hanchang: Yikes that's embarrassing, that's def the fix, thanks!

hanchang:
Closed in favor of #248

PR #104, opened 2018-05-29
fix News tab shows nothing until you click off and back #688079837795232

gduquesnay:


gduquesnay:
@hanchang First branch & pull request since we switched to Asana.  Not sure if I did the card# tagging properly.

PR #109, opened 2018-05-31
Reduce render-blocking scripts + convert large image to next gen format #687956476093810

gduquesnay:
https://app.asana.com/0/683768070860536/687956476093810


app/views/layouts/application.html.erb
```diff
@@ -23,11 +21,12 @@
       <%= yield %>
     </main>
     <%= render partial: "components/footer" %>
-    <%= render partial: "components/tracking_scripts" %>
+    <script src="/tradingview/charting_library.min.js"></script>
```
cryptoB0T: great idea to move scripts to the footer! another consideration is to make these asynchronous, rather than using synchronous script references. it would also be interesting to see the performance profile report from before and after. 

example: 
```js
  (function () {
        var s = document.createElement('script');
        s.type = 'text/javascript';
        s.async = true;
        s.src = /tradingview/charting_library.min.js';
        var x = document.getElementsByTagName('script')[0];
        x.parentNode.insertBefore(s, x);
    })();
```

app/views/pages/home.slim
```diff
@@ -8,7 +8,7 @@ section.phr.bg-white.tiber
               | Financial Intelligence for Cryptocurrencies
           .col-xs-12.col-md-5.col-lg-12
             .pl4-m.pl0-l.pr5-l.tc.pv3.pt0-m.pb0-l
-              img.max-w15e.max-w18e-s.max-w50e-m.w-100 src=image_path('ethereum-machine.png')
+              img.max-w15e.max-w18e-s.max-w50e-m.w-100 src=image_path('ethereum-machine.jpg')
```
cryptoB0T: I wonder how much faster this page would load if we were using webp. is there a scaleable way this can be done across the app, rather than manually?

PR #110, opened 2018-06-01
Fix  user registration page - style looks off #687956476093822

gduquesnay:
https://app.asana.com/0/683768070860536/687956476093822


hanchang:
This works again after the re-importing the prod database, my apologies!

PR #118, opened 2018-06-05
Newsfeed filter updates

MythicalFish:


app/javascript/App/components/NewsfeedPage/CoinDrawer.js
```diff
@@ -0,0 +1,52 @@
...
+const CoinDrawer = (props) => (
+  <div className="dn-l">
+    <button
+      className="btn btn-blue btn-xs"
+      onClick={() => props.toggleUI('coinDrawer')}
+    >
+      <Icon name="list" className="mr2" />
+      Coin list
+    </button>
+    <Animate
```
cryptoB0T: @MythicalFish great job here, I like the implementation!

app/javascript/App/components/NewsfeedPage/CoinDrawer.js
```diff
@@ -0,0 +1,52 @@
...
+      }}
+    >
+      {({ x }) => <CoinDrawerList {...props} x={x} />}
+    </Animate>
+  </div>
+)
+
+class CoinDrawerList extends Component {
+  onSwipeLeft = () => {
+    this.props.toggleUI('coinDrawer')
```
cryptoB0T: passing a string to the `toggleUI` method seems less than ideal. it might be interesting to consider passing the constructor name. the `toggleUI` method can handle matching the class to what it should be toggling. then invoking this method would be consistent across the application

MythicalFish: Don't quite understand this one, maybe you can show me some time

PR #122, opened 2018-06-06
Redirect logged in users from static homepage to Watchlist

gduquesnay:
https://app.asana.com/0/683768070860536/694173489004410


hanchang:
We're going to be removing the Watchlist since News has all of the same functionality.

PR #123, opened 2018-06-07
Newsfeed watchlist #698002095424671

MythicalFish:


app/assets/stylesheets/components/tooltip.scss
```diff
@@ -0,0 +1,84 @@
...
+  font-size: 0.8rem;
+  text-align: center;
+  font-style: normal;
+  position: absolute;
+  left: 50%;
+  max-width: 150px;
+  background: $bright-gray;
+  color: $white;
+  border-radius: 2px;
+  -webkit-border-radius: 2px;
```
cryptoB0T: is it a good idea to import the tachyons classes for border-radius and other properties, or is it better to just create custom ones? 

how do you determine whether to use the table of styles to look up what you need, or to just write it in?

app/controllers/api/newsfeed/coins_controller.rb
```diff
@@ -1,8 +1,12 @@
 class Api::Newsfeed::CoinsController < ApiController
 
   def index
-    @coins = Coin.order(:ranking).limit(20)
-    respond_success index_serializer(@coins)
+    coin_ids = Coin.order(:ranking).limit(20).ids
```
cryptoB0T: it looks like you're hardcoding the `.limit(20` in the controller? is this exposed in the endpoint, so the user can choose to request 50 coins if they needed to?

app/javascript/App/components/NewsfeedPage/CoinBody.js
```diff
@@ -1,64 +1,58 @@
...
+    return <LoadingIndicator className="overlay bg-white-70" />
+  return (
+    <div className="pa4">
+      <div className="flex justify-between items-center">
+        <a
+          href={`/coins/${coin.get('slug')}`}
+          className="f4 fw6 flex items-center color-inherit"
+        >
+          {coin.get('image_url') && (
+            <img className="w2e h2e mr3" src={coin.get('image_url')} alt="" />
```
cryptoB0T: this is just a comment: i don't feel it's worth the tradeoff, to adopt this framework that requires you add a bunch of abbreviated classes to do something simple. `className="w2e h2e mr3` just looks like greek, and the majority of people won't know what this means. to hit the ground running is impossible. they will need to have the API side by side and look up things. 

IMHO that in itself, is a good enough reason to avoid this like the plague. the vision is for the company to grow, and in that, the team will grow. think about what the community has adopted, rather than what has been successful for us. if we had jeremy ashkenas on our team, maybe we would get the most success using backbone. at the cost of making it harder for other people to work with and harder to attract talent. 

for the same reason we use react over ember. not that ember isn't well engineered, capable, or well documented. in fact many enterprise applications strongly evangelize ember. but it doesn't have a wide user base, and is harder to find ember developers. in the same vein, it's harder to find tachyon developers. they are a small subset of the css ecosystem.

app/javascript/App/components/NewsfeedPage/CoinBody.js
```diff
@@ -1,64 +1,58 @@
...
+      </div>
+      <div className="min-h12e flex items-center justify-center">
+        <div className="tc">
+          <div className="flex items-center">
+            <span className="f2">
+              <Currency>{coin.getIn(['market_info', 'price_usd'])}</Currency>
+            </span>
+            <span className="ml2">
+              <PercentageChange
+                number={coin.getIn(['market_info', 'percent_change_24h'])}
```
cryptoB0T: is using `number` better than just using `value`?

app/javascript/App/components/NewsfeedPage/CoinBody.js
```diff
@@ -1,64 +1,58 @@
...
+                className="b db"
+              />
+            </span>
+          </div>
+          <div className="dib ph2 pv1 bg-light-gray f6 mt2">
+            Market:&nbsp;
```
cryptoB0T: do you think it's better to use html entities or template literals here?

app/javascript/App/components/NewsfeedPage/CoinListHeader.js
```diff
@@ -0,0 +1,21 @@
...
+import Switch from '../Switch'
+
+const CoinListHeader = (props) => {
+  const { toggleUI, currentUI } = props
+  return (
+    <SectionHeader>
+      <div className="flex items-center">
+        <Switch
+          on={currentUI('watchingOnly')}
+          onChange={() => toggleUI('watchingOnly')}
```
cryptoB0T: any particular reason to return this invocation from an anonymous arrow function rather than just calling it?

app/javascript/App/components/NewsfeedPage/CoinListItem.js
```diff
@@ -1,33 +1,27 @@
...
 
-export default ({ coin, setActiveEntity, activeEntity }) => {
-  let klass = 'pa3 bb b--light-gray flex justify-between items-center pointer'
-  if (
-    activeEntity &&
-    activeEntity.type === 'coin' &&
-    activeEntity.id === coin.get('id')
-  )
-    klass += ' bg-foam'
+export default (props) => {
```
cryptoB0T: i think it's better to pass in just what we need. i don't like the idea of sending the ⛵️of props to descendants. let me know what you think!

app/javascript/App/components/NewsfeedPage/NewsList.js
```diff
@@ -13,7 +13,7 @@ const NewsItemList = (props) => {
         {newsItems
           .reverse()
           .map((newsItem) => (
-            <NewsItemListItemAnimated
+            <NewsListItemAnimated
```
cryptoB0T: nice one! i can't say i will miss you, Mr. `NewsItemListItem` 😢

app/javascript/App/components/NewsfeedPage/NewsListHeader.js
```diff
@@ -1,13 +1,14 @@
...
+import SectionHeader from './SectionHeader'
 import FilterPanel from '../FilterPanel'
 
-const ActionBar = (props) => {
+const NewsListHeader = (props) => {
   const { toggleUI, currentUI, coins, feedSources, activeFilters } = props
   return (
     <Fragment>
-      <div className="pa3 bb b--athens-dark flex justify-between items-center">
+      <SectionHeader>
```
cryptoB0T: this looks great, seems more readable to abstract this div into `SectionHeader`

app/javascript/App/components/WatchButton.js
```diff
@@ -3,52 +3,54 @@ import API from '../lib/localAPI'
...
+    if (!this.userIsLoggedIn) return
+    /* This component does its own API request & state management since it's
+    sometimes used on its own, and doesn't have access to Redux */
+    const { coinID: id, onWatch, onChange, setUser } = this.props
+    let { watching } = this.state
+    let params = { watchCoin: id }
+    if (watching) params = { unwatchCoin: id }
+    API.patch('/user', params).then(({ type, payload }) => {
+      console.log(params)
+      console.log(payload)
```
cryptoB0T: i think you left some logs 💩

app/javascript/App/components/WatchButton.js
```diff
@@ -3,52 +3,54 @@ import API from '../lib/localAPI'
...
-          </span>
-        )}
-      </button>
+      <div className="dib tooltipped">
+        {!this.userIsLoggedIn && <div className="tooltip">Login to watch</div>}
+        <button onClick={this.handleClick} className={btnClass}>
+          {watching ? (
+            <span>
+              <Icon name="star" solid className="mr1" />
+              Watching
```
cryptoB0T: let me know how you feel about creating a localization file and keeping our strings there. it's very possible we might want to support other languages and it makes the code cleaner. something feels wrong hardcoding strings in components! i just did it in another PR, and i feel like a 🤠

app/javascript/App/lib/redux/apiSagas.js
```diff
@@ -57,16 +57,15 @@ function handleError({ error }) {
...
-    console.log(response.message)
-  }
-  let payload
-  if (response.payload) {
-    payload = response.payload
-  } else {
-    payload = response
+function* handleResponse(callback, response, url) {
+  const { type, payload, message } = response
+  if (!response) {
```
cryptoB0T: do all these conditions get logged in rollbar? is there another way to handle these cases besides logging to the console?

cryptoB0T:
great job, @MythicalFish! most of these are just comments, don't feel obliged to change every line. looks mostly good to me, use your discretion.

app/assets/stylesheets/components/tooltip.scss
MythicalFish: Yeah there's no set rule here. For the border radius we should indeed have a single value that we use everywhere. Tachyons itself doesn't provide though, it's totally agnostic

app/controllers/api/newsfeed/coins_controller.rb
MythicalFish: Yep 20 is what we want for now, easy to change though

app/javascript/App/components/NewsfeedPage/CoinBody.js
MythicalFish: I'm a big fan of functional CSS frameworks like Tachyons or BassCSS, and the learning curve isn't steep I think. What I do is keep the table of styles open in a tab and search for the utility I need.
This page makes a good case: https://github.com/dwyl/learn-tachyons#30-second-summary

I'd be open to using something like Styled Components, but only because it would indeed allow other devs to get started more quickly.

app/javascript/App/components/NewsfeedPage/CoinBody.js
MythicalFish: Yep, because of reasons.

MythicalFish: (No, we can change this ;))

app/javascript/App/components/NewsfeedPage/CoinBody.js
MythicalFish: No strong opinion here. Probably template literals are nicer

app/javascript/App/components/NewsfeedPage/CoinListHeader.js
MythicalFish: Yeah, wouldn't work otherwise right?
If toggleUI didn't have any arguments then we could indeed just do `onchange={toggleUI}`, but since we do then it needs to be in a function to avoid it being called on render

app/javascript/App/components/NewsfeedPage/NewsList.js
MythicalFish: Haha <3

app/javascript/App/components/NewsfeedPage/CoinListItem.js
MythicalFish: I definitely favor passing in everything, in past projects I spent (wasted) a lot of time on shuffling individual props around. But if you think that we'll see performance issues in future then we can of course. I guess for small, frequently rendered components like this it makes sense.

app/javascript/App/components/WatchButton.js
MythicalFish: T____T

app/javascript/App/components/WatchButton.js
MythicalFish: Yeah definitely onboard

app/javascript/App/lib/redux/apiSagas.js
MythicalFish: No they don't, will make an Asana ticket for that

MythicalFish:
@cryptoB0T Shall we merge this? Not sure what the protocol is here, it needs to be squashed right?

app/javascript/App/components/NewsfeedPage/CoinListHeader.js
cryptoB0T: well, you could use your good friend, bind 😄 

`toggleUI.bind(this, 'watchingOnly')`

app/javascript/App/components/NewsfeedPage/CoinListItem.js
cryptoB0T: here's a thread that discusses exactly this, albeit with very little fanfare: 

https://stackoverflow.com/questions/43532684/in-react-is-it-a-good-practice-to-pass-all-the-props-from-parent-to-child-compo

cryptoB0T:
🚢

app/javascript/App/components/NewsfeedPage/CoinBody.js
hanchang: @cryptoB0T For what it's worth, I started off a disbeliever but kept an open mind, and less than a week into development I was already reasonably fluent in Tachyons.

It's a lot more flexible than something like Bootstrap and the learning curve is pretty shallow - although the beginning feels very "Greek" in that it's a jumble of letters, the paradigm is quite straightforward to pick up and the versatility started feeling very powerful.

Think of it like `vim` for frontend design; the initial learning curve sucks but once you understand the paradigm, it's pretty cool to be able to do compositions and the expressiveness is much nicer than other more rigid frontend frameworks.

PR #132, opened 2018-06-13
Newsfeed infinite scroll #695422732378435

jacobi-78:


MythicalFish:
Not quite working yet but overall good work :)

jacobi-78:
Leaving a quick note that I verified the current deployed version works.

app/javascript/App/components/NewsfeedPage/NewsList.js
```diff
@@ -1,31 +1,84 @@
...
+  componentWillUnmount() {
+    this.unmountOnScrollHandler()
+  }
+
+  mountOnScrollHandler() {
+    if (window.isMobile || window.isTablet) {
+      const throttled = _.throttle(this.props.onScrollNewsFeedMobile, 1000)
+      $(window).scroll(throttled)
+    } else {
+      const throttled = _.throttle(this.props.onScrollNewsFeedDesktop, 1000)
```
MythicalFish: I'd rather move the distance/buffer detection here rather than in the action creator (and not dispatch {type:null})

app/javascript/App/components/NewsfeedPage/NewsList.js
```diff
@@ -1,31 +1,84 @@
...
-          {newsItems.map((newsItem) => (
-            <NewsListItemAnimated
-              key={newsItem.get('id')}
-              {...this.props}
-              newsItem={newsItem}
-              height={itemHeight}
-            />
-          ))}
+        <div id="news-feed"
+          className={'flex-auto relative overflow-y-' + (isMobile ? 'hidden' : 'auto')}>
```
MythicalFish: You can use Tachyons utilities with ns/m/l extension for applying a class to a screen size.
Eg. `overflow-y-hidden overflow-y-auto-m`
But in this case I think it should be set to auto regardless of the screen size

app/javascript/App/containers/newsfeed/actions.js
```diff
@@ -4,7 +4,35 @@ import { createEntityActions, createFilterActions } from '../../lib/redux'
 const entityActions = createEntityActions(namespace)
 const filterActions = createFilterActions(namespace)
 
+const scrollBottomToReducer = (scrollBottom) => {
+  if (!scrollBottom) return { type: null }
```
MythicalFish: As I mentioned above, I'd rather not dispatch { type: null }

app/javascript/App/containers/newsfeed/actions.js
```diff
@@ -4,7 +4,35 @@ import { createEntityActions, createFilterActions } from '../../lib/redux'
...
+
+  return { type: 'FETCH_MORE_NEWS_FEED' }
+}
+
+const onScrollNewsFeedMobile = (e) => {
```
MythicalFish: I think it would be nice to have a single action creator for desktop & mobile, which does the mobile detection (rather than doing that from the component)

app/javascript/App/containers/newsfeed/reducer.js
```diff
@@ -3,7 +3,18 @@ import { namespace, filterList } from './constants'
 
 const initialState = {}
 
-const newsfeedReducer = (state, action) => state
+const newsfeedReducer = (state, action) => {
```
MythicalFish: 🎉

app/javascript/App/containers/newsfeed/sagas.js
```diff
@@ -107,3 +110,21 @@ function* newsitemParams() {
   }
   return params
 }
+
+function* onScrollingToBottom(action) {
```
MythicalFish: 🎉

jacobi-78:
Do not test or review yet, I'm still testing the feature.

jacobi-78:
The functionality works now, feel free to test it.

MythicalFish:
Fixed the merge conflicts.
Had quite some confusion because in my local persisted state the `isLoading` for `newsFeed` wasn't set, and caused the infinite scroll to not work.

But yeah, looks good! 🚀

jacobi-78:
There's still an issue with redux caching which I will push shortly -- it works initially, but then breaks. Can you take off your last commit? I want to test some changes I made and then re-merge your work.

jacobi-78:
Fixed the final issue with redux caching, should be good to go! 👍

PR #133, opened 2018-06-13
On-off toggle button for price charts #695422732378435

jacobi-78:
* show / hide pop-up bubbles on chart related to news stories


MythicalFish:
Hi @jacobi-78 
I can't see what's been done here?
It's also a bit difficult because of all the linting 
Ideally only files you've worked on should be linted

PR #134, opened 2018-06-13
Features/newsfeed search/#701168722099913

MythicalFish:


app/javascript/App/components/NewsfeedPage/CoinList.js
```diff
@@ -1,29 +1,58 @@
...
 import CoinListItem from './CoinListItem'
 import LoadingIndicator from '../LoadingIndicator'
+import coinSearchProvider from '../../containers/coinSearch'
 
-export default (props) => {
-  const { isLoading, setActiveEntity, currentUI, isWatching } = props
-  const onClick = (coin) =>
-    setActiveEntity({
+class CoinList extends Component {
+  onClickCoin = (coin) =>
```
cryptoB0T: IMHO this function could have a more meaningful name. instead of calling it what happens, 'on click coin', it may be more readable calling it by what it is going to do, 'set active'. then the user can derive what it's doing and see see this in the callstack. 

so instead of looking like: 
```
onClick={onClickCoin}
onClickCoin = (coin) => setActiveEntity(coin)
```

it would look like this: 
```
onClick={setActive}
setActive = (coin) => setActiveEntity(coin)
```

then later you have another handler, that's preparing the ui by clearing the search and invoking toggleUI. 

instead of looking like: 

```
onClickNewCoin = (coin) => {
  this.onClickCoin(coin)
```

this could look like: 

```
newCoinHandler = (coin) => {
  this.setActive(coin)
```

app/javascript/App/components/NewsfeedPage/CoinList.js
```diff
@@ -1,29 +1,58 @@
...
+    }
+    return (
+      <Fragment>
+        <CoinListHeader {...this.props} />
+        <div className="flex-auto relative overflow-y-auto">
+          {isLoading('coins') && (
+            <LoadingIndicator className="overlay bg-white-70" />
+          )}
+          {currentUI('coinSearch') &&
+            searchedCoins.size > 0 && (
```
cryptoB0T: any values greater than zero will always evaluate to true.

app/javascript/App/containers/coinSearch/sagas.js
```diff
@@ -10,7 +11,7 @@ function* searchCoins({ searchText, searchOpts, namespace }) {
   if (searchText.length < 2) return
   yield apiSagas.get(
     '/coins.json',
-    Object.assign({ q: { name_or_symbol_cont: searchText } }, searchOpts),
+    _.merge({ q: { name_or_symbol_cont: searchText } }, searchOpts),
```
cryptoB0T: why did you choose to use `_.merge` instead of `Object.assign`?

app/javascript/App/components/NewsfeedPage/CoinList.js
MythicalFish: Makes sense :)

app/javascript/App/components/NewsfeedPage/CoinList.js
MythicalFish: I think having `> 0` is a bit more human-friendly

app/javascript/App/containers/coinSearch/sagas.js
MythicalFish: `Object.assign` is shallow

PR #136, opened 2018-06-13
Feature/feedsource updates/#707415781908419

MythicalFish:


db/migrate/20180613155411_change_feed_source_site_url_to_hostname.rb
```diff
@@ -0,0 +1,8 @@
+class ChangeFeedSourceSiteUrlToHostname < ActiveRecord::Migration[5.1]
```
hanchang: @MythicalFish Could you please provide your thought process for changing the schema?

This will require propagating a lot of changes on the data engineering side, so I'd strongly prefer to avoid this situation if at all possible!

MythicalFish: Fixed the migration as discussed :)

hanchang:
We're clear from a data engineering perspective.

PR #139, opened 2018-06-14
wrap text on newsitem in mobile #708844042045442

cryptoB0T:


MythicalFish:
@cryptoB0T Had a figure out how to fix this myself so I thought I may as well just push what I'd done. :) What you had before caused linebreaks in the wrong places

PR #142, opened 2018-06-15
Remove paper_trail temporarily until we figure out how to only log ma…

gduquesnay:
…nual admin user modifications #695422732378416


hanchang:
Deployment Notes: Migration required.

PR #144, opened 2018-06-15
Integrate NewsTagger with Platform on Staging

gduquesnay:
https://app.asana.com/0/683768070860536/705852416652323


gduquesnay:
@hanchang Couldn't think of a way to test this in dev  ¯\_(ツ)_/¯.   Hope it works.

gduquesnay:
@hanchang Just updated this to not fail if NEWS_TAGGER_ENDPOINT is not configured.  So it's easier to work with in other devs local and so we don't have to worry about the timing of deploying it.

PR #155, opened 2018-06-19
Bugfix/news panel look feel/#710451747386678

cryptoB0T:
laptop
image

tablet
localhost_3000_news_q categories 0 partnerships q categories 1 regulatory ipad

mobile
localhost_3000_news_q categories 0 partnerships q categories 1 regulatory pixel 2 xl

### DIFF
--- a/app/javascript/App/components/FilterPanel/FilterComponent.js
+++ b/app/javascript/App/components/FilterPanel/FilterComponent.js
@@ -22,7 +22,7 @@ const FilterComponent = (props) => {
         <Categories {...props} />
       </div>
       <div className="pv4 bb b--geyser" style={newsFeedStyle ? {paddingTop:'.5rem', paddingBottom:'1.5rem'}: ''}>
-        <h4 className="mb2" style={{margin:'15px 0'}}>Sources</h4>
+        <h4 className="mb2" style={{margin:'15px 0', fontWeight:'bold'}}>Sources</h4>
         <FeedSources {...props} />
       </div>
     </Fragment>
--- a/app/javascript/App/components/FilterPanel/FilterComponent.js
+++ b/app/javascript/App/components/FilterPanel/FilterComponent.js
@@ -1,19 +1,31 @@
-import React from 'react'
+import React, { Fragment } from 'react'
 import Types from 'prop-types'
 import components from './filterComponents'
+import MarketMoving from './filterComponents/MarketMoving'
+import Categories from './filterComponents/Categories'
+import FeedSources from './filterComponents/FeedSources'
 
 const FilterComponent = (props) => {
-  const { filter } = props
+  const { filter, newsFeedStyle } = props
   const Component = components[filter.get('key')]
   if (!Component) {
-    console.error(`Component not found for "${filter.get('key')}"`)
     return null
   }
   return (
-    <div className="pv4 bb b--geyser">
-      <h4 className="mb2">{filter.get('label')}</h4>
-      <Component {...props} />
-    </div>
+    <Fragment>
+      <div className="pv4 bb b--geyser" style={newsFeedStyle ? {paddingTop:'.5rem', paddingBottom:'1.5rem'}: ''}>
+        <h4 className="mb2">Market moving</h4>
+        <MarketMoving {...props} />
+      </div>
+      <div className="pv4 bb b--geyser" style={newsFeedStyle ? {paddingTop:'.5rem', paddingBottom:'1.5rem'}: ''}>
+        <h4 className="mb2" style={{margin:'15px 0'}}>Categories</h4>
+        <Categories {...props} />
+      </div>
+      <div className="pv4 bb b--geyser" style={newsFeedStyle ? {paddingTop:'.5rem', paddingBottom:'1.5rem'}: ''}>
+        <h4 className="mb2" style={{margin:'15px 0'}}>Sources</h4>
+        <FeedSources {...props} />
+      </div>
+    </Fragment>
   )
 }
 
--- a/app/javascript/App/components/FilterPanel/Layout.js
+++ b/app/javascript/App/components/FilterPanel/Layout.js
@@ -2,7 +2,7 @@ import React from 'react'
 import Icon from '../Icon'
 
 const Layout = (props) => {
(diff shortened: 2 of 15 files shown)


app/javascript/App/components/FilterPanel/filterComponents/Categories.js
```diff
@@ -1,17 +1,17 @@
 import React from 'react'
 import Type from 'prop-types'
-import ItemSelector from '../../ItemSelectorAlt'
+import ItemSelectorCategory from '../../ItemSelectorCategory'
```
MythicalFish: I think "Category" is too specific for it to live in the root of the /components/ dir

app/javascript/App/components/FilterPanel/index.js
```diff
@@ -30,16 +30,18 @@ class FilterPanel extends Component {
     const pProps = { ...props, filterData: normalizeFilterData(filterData) }
     const { applyFilters, resetFilters } = this
     return (
-      <Layout {...{ ...props, applyFilters, resetFilters }}>
+      <Layout {...{ ...props, applyFilters, resetFilters }} newsFeedStyle>
         {filterList.map((filter, index) => {
           if (filter.get('key') === 'coins') return null // Temp fix for hiding coins
+          if (filter.get('key') === 'keywords') return null
```
MythicalFish: We definitely want to hide this? I seem to remember hearing that it would be nice to have

app/javascript/App/components/ItemSelectorAlt.js
```diff
@@ -2,19 +2,25 @@ import React, { Component } from 'react'
...
+  width: 'auto',
+  marginRight: '.5rem'
+}
+
 export default class ItemSelectorAlt extends Component {
   selectedItems = () => this.props.selectedItems || []
   isSelected = (item) => {
-    const selected = this.selectedItems().map((item) => JSON.stringify(item))
-    return selected.includes(JSON.stringify(item))
+    const selected = this.selectedItems().feedSources && this.selectedItems().feedSources.length && this.selectedItems().feedSources.map((item) => JSON.stringify(item))
```
MythicalFish: feedSources is too specific, this component should be reusable

app/javascript/App/components/FilterPanel/FilterComponent.js
```diff
@@ -1,19 +1,31 @@
...
+import MarketMoving from './filterComponents/MarketMoving'
+import Categories from './filterComponents/Categories'
+import FeedSources from './filterComponents/FeedSources'
 
 const FilterComponent = (props) => {
-  const { filter } = props
+  const { filter, newsFeedStyle } = props
   const Component = components[filter.get('key')]
   if (!Component) {
-    console.error(`Component not found for "${filter.get('key')}"`)
```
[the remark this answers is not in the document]
cryptoB0T: I've been using flow types in my recent projects, type-safety is great. i don't know how i feel about checking for the existence of a string in the list object, and logging an error to the console. thanks for the suggestion but i think it's better this way. with this PR you don't have everything passing through this component anymore.

app/javascript/App/components/FilterPanel/FilterComponent.js
```diff
@@ -1,19 +1,31 @@
...
     return null
   }
   return (
-    <div className="pv4 bb b--geyser">
-      <h4 className="mb2">{filter.get('label')}</h4>
-      <Component {...props} />
-    </div>
+    <Fragment>
+      <div className="pv4 bb b--geyser" style={newsFeedStyle ? {paddingTop:'.5rem', paddingBottom:'1.5rem'}: ''}>
```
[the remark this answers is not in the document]
cryptoB0T: in this case i specifically didn't want to create an extraneous wrapping tag, don't need to assign a class here. if at a later time we determine we need a wrapping tag to assign a class, we can change it then.

app/javascript/App/components/FilterPanel/Layout.js
```diff
@@ -20,8 +20,14 @@ const Layout = (props) => {
           <button
             className="btn btn-white btn-xs"
             onClick={() => disableUI('filterPanel')}
+            style={{
```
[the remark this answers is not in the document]
cryptoB0T: there was a lot of discrepancies on how this button rendered in the browser and the visual design comps provided. at this point the risk of introducing a visual regression due to modifying an existing button used in other parts of the site, trumps the desire for reusability. 

our objective at this point 
1 achieve near-perfection in implementing the interface so it looks like the comps
2 build reusable components we can compose together 

when the smoke clears we can revisit how to make this reusable. at the moment a correct implementation of this button does not exist.

app/javascript/App/components/FilterPanel/filterComponents/Categories.js
cryptoB0T: i'll reply to this one in comments of the PR

app/javascript/App/components/FilterPanel/index.js
cryptoB0T: it's great that you remember hearing something about how the requirements may have evolved. i think we should move to be adamant about having a single source of truth. that could very well be the design comps. 

this is non-existent in the comps so i took it out. but it's just one line to put it back, when we need it!

app/javascript/App/components/ItemSelectorAlt.js
cryptoB0T: i'll reply to this one in comments of the PR

cryptoB0T:
there seems to be some differences in how we code. your approach is like the generalist: make your code supremely reusable so that it can be used for anything. your belief is that this results in more code reuse.

i believe we achieve maximum productivity in following the agile manifesto. write code that does exactly what i need it to do. later if i find i need to reuse it, i can just change the names later. 

this way works faster, works better, introduces less bugs, and saves a lot of time, leaving us to have more time to implement features rather than over-architect solutions. i have done this with teams where we had excess of 95% test coverage. changing a name is a single grep command. time spent debugging implementations and fixing bugs due to ambiguous names is hard to quantify. 

thanks for reviewing my code, and i'm always happy to change anything to make it better. the comments in this PR are mostly opinions.

PR #157, opened 2018-06-20
Feature/coin filtering/#713015422631310

MythicalFish:
https://app.asana.com/0/683768070860536/713015422631310


app/controllers/api/news_items_controller.rb
```diff
@@ -6,11 +6,12 @@ def index
     q = params[:q] || {}
 
     @news_items = NewsItem.all
-
-    if q[:coinIDs]
-      news_item_ids_for_coin_filter = NewsCoinMention.where(coin_id: q[:coinIDs]).pluck(:news_item_id)
-      @news_items = @news_items.where(id: news_item_ids_for_coin_filter)
-    end
+    
+    coin_ids = q[:coinIDs]
```
cryptoB0T: will this result in us having brackets `[]` in our urls?

app/assets/stylesheets/components/forms.scss
```diff
@@ -57,12 +57,13 @@ textarea {
...
     -webkit-box-shadow: none;
     background: transparent;
+    line-height: 1.5;
   }
 }
 .icon-input {
   position: relative;
   input {
-    padding-left: 3rem;
+    padding-left: 3rem !important;
```
cryptoB0T: just a question: was there a reasonable way to do this without having to resort to using `!important`?

app/javascript/App/components/FilterTags.js
```diff
@@ -1,24 +1,24 @@
...
       {activeFilters.map((filter, i) => (
-        <div
-          key={i}
-          className="tag-alt2"
-          style={{padding:'0 .25rem 0 .25rem'}}
-        >
-          <span className="ml2">
-            <FilterButtonValue filter={filter} />
-          </span>
+        <div key={i} className="tag-alt2">
```
cryptoB0T: i think using the index as a key is not recommended, at one point in time React would throw a warning. is there a unique identifier we can use? in general, single letter variable names are frowned-upon.

app/javascript/App/components/NewsfeedPage/SearchCoins.js
```diff
@@ -0,0 +1,78 @@
...
+class SearchCoins extends Component {
+  selectedCoins = () => {
+    const { activeFilters } = this.props
+    const filter = activeFilters.find((filter) => filter.get('key') === 'coins')
+    if (!filter) return []
+    return filter.get('value').toJS()
+  }
+  handleSearchInput = (value) => {
+    let { searchCoins } = this.props
+    const name_not_in = this.selectedCoins()
```
cryptoB0T: the convention in js and in our app is mostly camel case. was there a reason you chose to use snake case here?

MythicalFish: Yes in JS I always cameCase, however in Ruby it's snake case, and in this instance it's easier to pass in the param that the controller needs. Otherwise if the param name changes, it needs to be changed in 2 places instead of 1

MythicalFish:
@hanchang Resolved the issues here, ready to merge unless I've missed something?

PR #158, opened 2018-06-20
Show "Watch" button not just the star #708230108698466

jacobi-78:


app/javascript/App/components/NewsfeedPage/CoinBody.js
```diff
@@ -24,7 +24,7 @@ const CoinBody = (props) => {
           <span className="ml2">({coin.get('symbol')})</span>
         </a>
         <div>
-          <WatchButton {...props} coin={coin} />
+          <WatchButton {...props} coin={coin} hasText={true} />
```
cryptoB0T: omitting the value of an attribute will evaluate in descendant props, as true 

another way to write this:

```
<WatchButton {...props} coin={coin} hasText />
```

PR #160, opened 2018-06-20
Display newscategories on admin newsitem show page from master/695422732378424

gduquesnay:


gduquesnay:
@hanchang as requested branched from master and cherry-picked.

PR #196, opened 2018-07-03
#718238341407898 Add javascript to force video loading and force play…

anthotsang:
https://app.asana.com/0/709175359523131/718238341407898

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -243,6 +243,28 @@
       }
     }
     tryHook();
+
+    $(document).ready(function() {
+				$('video.product-video').each(function(index) {
+					var video = $(this);
+					
+					video.on('error', function(e) {
+						console.log('error', e);
+					})
+
+					video.one('canplaythrough', function() {
+						$('body').on('click', function() {
+							$('video.product-video')[index].play();
+						});
+					});
+
+					video.on('ended', function() {
+						$('video.product-video')[index].play();
+					})
+
+					$('video.product-video')[index].load();
+				});
+			});
   </script>
   <button type="button" id="track-submit" style="display: none;"></button>
 </body>


public/index.html
```diff
@@ -243,6 +243,28 @@ <h1>Do you want early access to CoinFi News?</h1>
       }
     }
     tryHook();
+
+    $(document).ready(function() {
+				$('video.product-video').each(function(index) {
```
hanchang: @anthotsang Sorry for forgetting to mention during onboarding - the team has agreed to use spaces instead of tabs for indents.

public/index.html
```diff
@@ -243,6 +243,28 @@ <h1>Do you want early access to CoinFi News?</h1>
...
+					var video = $(this);
+					
+					video.on('error', function(e) {
+						console.log('error', e);
+					})
+
+					video.one('canplaythrough', function() {
```
hanchang: Should this be `video.on('canplaythrough', ...` ?

anthotsang: That event seems to fire continuously and I only need it to fire once so it's set to one instead of on. I guess I should also set the inner event listener to one.

hanchang: Gotcha, thanks for the explanation!

anthotsang:
@hanchang changes made. 
Is there an actual block to merging or do we just wait for a it's okay comment?

hanchang:
Thanks, indentation is still a little weird on line 248 but otherwise LGTM!

Generally we like to have at least one person sign off on a PR before merging, and the preference is for the team member who created the PR to do the honors.

We also prefer "Squash and merge" to keep the commit history a bit cleaner.

PR #198, opened 2018-07-04
Update about page roadmap to include 2019q4 and other minor updates

anthotsang:
Haven't tested locally because my environment isn't fully set up. It seems pull requests spin up a heroku instance? It's also just html changes.


hanchang:
@anthotsang Yup, PRs spin up a Heroku Review App.

PR #213, opened 2018-07-09
[WIP] Chore/clean up newsfeed query params/#735778520910407

cryptoB0T:


cryptoB0T:
@hanchang passing timestamp doesn't seem to work with infinite scroll..

cryptoB0T:
@hanchang viewing watchlist in left rail causes page refresh and doesn't work in this branch

PR #220, opened 2018-07-13
Add Launch Darkly. Hide news page behind 'news' feature flag.

anthotsang:
https://app.asana.com/0/709175359523131/742654163136024

I'm just initializing as a global variable right now. LD needs to be a singleton but it seems senseless to create a singleton class where the only thing exposed is the one variable. I guess the user helper could be a definition?

Also, not sure what to do about Heroku Env Vars


anthotsang:
@hanchang 

Please review when you get a chance. It seems has_news_feature? and get_ld_user should be refactored and placed somewhere else, but I'm not really sure what would be best/recommended.

Also, I tried adding the env vars to heroku but every time it build it seems to clobber them, but I'm not sure where it's getting them from.

app/controllers/api/coins_controller.rb
```diff
@@ -1,18 +1,26 @@
 class Api::CoinsController < ApiController
 
   def index
-    query = params[:q] || {}
-    if params[:exclude_watched]
-      query[:id_not_in] = current_user.watchlist.coin_ids
+    if (!has_news_features?)
```
hanchang: @anthotsang Could we use `return respond_unfound if !has_news_features?` to keep the rest of the code untouched for easier diff?

app/controllers/application_controller.rb
```diff
@@ -17,4 +17,19 @@ def default_url_options
     return {} if I18n.locale == I18n.default_locale
     { locale: I18n.locale }
   end
+
+  def has_news_feature?
```
hanchang: @anthotsang Let's put the `has_news_feature?` and `get_ld_user` method definitions within a `protected` section instead of `private` to expose them to child controllers and keep things DRY.

app/controllers/pages_controller.rb
```diff
@@ -29,4 +33,19 @@ def member_pages
     %w[]
   end
 
+  def has_news_feature?
```
hanchang: These won't be necessary after the `ApplicationController` refactor suggestion to move from `private` to `protected`.

config/initializers/launch_darkly.rb
```diff
@@ -0,0 +1 @@
+$ld_client = LaunchDarkly::LDClient.new(ENV.fetch("LAUNCHDARKLY_SDK_KEY"))
```
hanchang: I know the `$ld_client` global variable name is what their docs suggest, but I'm thinking it might be easier for future devs to pick up what's going on if we rename this to something like `$launch_darkly` instead?

app/controllers/api/coins_controller.rb
hanchang: If that (or some variation thereof) works, please go ahead and implement across all invocations as well, thanks!

hanchang:
Great start @anthotsang!

I think `ApplicationController` is the best place to put the `has_news_feature?` and `get_ld_user` methods, just need to put them in a `protected` section instead of `private` to expose them to child controllers in the inheritance hiearchy.

For the env var issue on Heroku, what we’re missing is the LaunchDarkly env vars added to the `coinfi-staging-eu` app since that’s the parent of every review app. We’ll also need to add the new env vars to the `app.json` file to tell Heroku to copy those env vars over from staging to new review apps.

https://devcenter.heroku.com/articles/app-json-schema

Lemme know if you have any questions on these two suggestions!

config/initializers/launch_darkly.rb
hanchang: This is still missing the `$launch_darkly` definition?

PR #231, opened 2018-07-16
move categories above fold #744653817821601

cryptoB0T:


hanchang:
Closed in favor of #233

PR #239, opened 2018-07-19
Feature/lint staged/#749359417073858

cryptoB0T:
- [x] precommit hook that will run linter before commit files 
- [ ] run prettier on most of the files

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,3 +1,6 @@
+-# Don't tidy legacy JS
+-app/assets/**/*.js
+
 dist
 node_modules
 coverage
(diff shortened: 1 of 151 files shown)


hanchang:
@cryptoB0T Awesome stuff - let's hold off on this until we launch the private beta and take that reprieve to do this in one shot. Right now there's too much going on to make this massive change viable.

cryptoB0T:
@hanchang I went ahead and reverted the single commit hash that lints all the files. then this becomes really small. the downside is that when you touch a file, it will require that you lint the file then.

cryptoB0T:
@hanchang this should have been there from the start. let's try to get this in asap because this will relieve us from policing code style in PRs

PR #240, opened 2018-07-19
Improve Newsfeed Query Performance

hanchang:
https://app.asana.com/0/683768070860536/744567967891137

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,7 +32,6 @@ gem 'jbuilder', '~> 2.5'
 # Use Capistrano for deployment
 # gem 'capistrano-rails', group: :development
 
-gem 'active_record_union'
 gem 'acts-as-taggable-on'
 gem 'administrate'
 gem 'administrate-field-belongs_to_search'
--- a/Gemfile
+++ b/Gemfile
@@ -95,6 +95,7 @@ group :development do
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
   gem 'spring'
   gem 'spring-watcher-listen', '~> 2.0.0'
+  gem 'bullet'
 end
 
 group :production do
--- a/app/controllers/api/news_items_controller.rb
+++ b/app/controllers/api/news_items_controller.rb
@@ -11,11 +11,23 @@ class Api::NewsItemsController < ApiController
     coin_ids = Coin.where(name: q[:coins]).pluck(:id) if q[:coins]
     coin_ids = Coin.top(20).ids unless coin_ids
 
-    @news_items = NewsItem.published.joins(:news_coin_mentions).where(news_coin_mentions: { coin: coin_ids })
+    @news_items = NewsItem.all
 
-    if q[:coins].blank? && q[:feedSources].blank?
-      # Only show NewsItems from General FeedSources when no coins are specifically selected.
-      @news_items = NewsItem.general.published.joins(:news_coin_mentions).where(news_coin_mentions: { coin: coin_ids })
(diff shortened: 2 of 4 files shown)


Gemfile
```diff
@@ -95,6 +94,7 @@ group :development do
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
   gem 'spring'
   gem 'spring-watcher-listen', '~> 2.0.0'
+  gem 'bullet'
```
justin808: Good call on this one. We've been using it for a long time on HawaiiChee.com.

app/controllers/api/news_items_controller.rb
```diff
@@ -11,28 +11,31 @@ def index
...
     end
 
-    if q[:categories].present?
-      category_ids = NewsCategory.where(name: q[:categories]).ids
-      news_item_ids_for_category_filter = NewsItemCategorization.where(news_category_id: category_ids).pluck(:news_item_id)
-      if news_item_ids_for_category_filter.present?
-        @news_items = @news_items.where(id: news_item_ids_for_category_filter)
-      end
+    # Showing default coins
+    if coin_ids == q[:coinIDs]
```
justin808: Can you be sure that that the coinIDs will always be sorted and comparable with == ? As in you might be comparing an array to a string, etc.

app/controllers/api/news_items_controller.rb
```diff
@@ -11,28 +11,31 @@ def index
...
+      coin_news_item_ids = NewsCoinMention.where(coin_id: coin_ids).pluck(:news_item_id)
```
justin808: This runs an extra query.

Did you try doing this without an extra query and just having the whole query as part of the or clause with a join?

app/controllers/api/news_items_controller.rb
```diff
@@ -11,28 +11,31 @@ def index
...
       if feed_source_ids.present?
         @news_items = @news_items.where(feed_source_id: feed_source_ids)
       end
-    else
-      @news_items = @news_items.general
+    end
+
+    if q[:categories].present?
+      category_ids = NewsCategory.where(name: q[:categories]).ids
+      news_item_ids_for_category_filter = NewsItemCategorization.where(news_category_id: category_ids).pluck(:news_item_id)
```
justin808: Here are a couple extra queries...You might be able to combine these.

app/controllers/api/news_items_controller.rb
```diff
@@ -11,28 +11,31 @@ def index
...
+      if news_item_ids_for_category_filter.present?
+        @news_items = @news_items.where(id: news_item_ids_for_category_filter)
```
justin808: So if nothing matches for `news_item_ids_for_category_filter` then you don't want any filtering by this?

app/controllers/api/news_items_controller.rb
hanchang: Good instincts, but yes, since we're basically trying to detect the assignment on line 10.

app/controllers/api/news_items_controller.rb
hanchang: I'm aware, and I did attempt to do this without an extra query but this is what's causing the performance issues. In the interest of saving dev time I just opted for the ugly IDs clause temporarily.

app/controllers/api/news_items_controller.rb
hanchang: Same situation as above.

PR #245, opened 2018-07-19
Create calendar events/742654163136026

anthotsang:
https://app.asana.com/0/709175359523131/742654163136026


anthotsang:
> Probably need two sources, one for the URL and another for a screenshot or something along those lines?

I guess I was assuming that a screenshot would get uploaded somewhere and then we could still just use URL to link to the screenshot. What are your thoughts? Better to keep it separate?

hanchang:
@anthotsang Ah sorry just saw this, apologies for being unclear, this is what I had in mind:

The `source_url` is a string representing the URL of the web page where the information was detected by the curator user.
The `screenshot_url` is a string representing the URL of the uploaded screenshot of the `source_url` that the curator user took.

Hope that clears things up!

db/migrate/20180715114637_create_calendar_events.rb
```diff
@@ -0,0 +1,28 @@
+class CreateCalendarEvents < ActiveRecord::Migration[5.1]
+  def change
+    create_table :calendar_events do |t|
+      t.references :creator
```
hanchang: Sorry, I think this should still be `t.references :user` because otherwise I don't think Rails is smart enough to connect the dots. But we can just put a comment in the model in the CalendarEvents has_one :user association to note that the user refers to the creator of the calendar event.

hanchang:
LGTM, I'm not 100% sure if the renaming of the user reference to `creator` will work out of the box - if it does, feel free to squash and merge as is!

db/migrate/20180715114637_create_calendar_events.rb
anthotsang: In the end I couldn't guarantee it was working properly, still not sure what the easiest way to verify that the association was working properly. At an SQL level, I believe my foreign key was correct, but I had no idea how to get the rails model to understand it.

PR #250, opened 2018-07-20
move script into js, since parsed as string in template

cryptoB0T:
@hanchang do you know the history of this code? what is this js doing if this is a server rendered form?


app/views/users/registrations/new.html.erb
```diff
@@ -20,11 +20,3 @@
     </div>
   </div>
 </div>
-
-<script>
```
hanchang: We need to keep this script, otherwise users will double click on the Signup button and see an error message saying an account already exists.

hanchang:
Cool, this works - we'll need to remember to rip this out once we switch over to the new onboarding flow.

PR #252, opened 2018-07-21
Fix password reset page so it's like the login page.

anthotsang:
https://app.asana.com/0/683768070860536/751599389132228


hanchang:
💯

PR #267, opened 2018-07-28
Fix overagressive linting. ES lint should only be run on javascript-l…

anthotsang:
…ike files.


hanchang:
Thanks for the cleanup!

PR #268, opened 2018-07-28
Create calendar component and associated api

anthotsang:


app/controllers/pages_controller.rb
```diff
@@ -4,7 +4,7 @@ def show
     render_404 unless page_known?
     redirect_to('/login') && return if member_page? && !current_user
     @body_id = "#{@page}-page"
-    if @page == 'news' && !has_news_feature?
+    if !has_feature?
```
hanchang: I think one refactor we need to do later is to separate each page into its own controller, this is getting a bit dirty. Thanks for sticking with the existing paradigm, just flagging that we'll want to change this shortly.

hanchang:
🥇 This is fantastic progress!

PR #273, opened 2018-07-31
Exchange Listings Page

justin808:


app/javascript/App/components/ExchangeListingsPage/LayoutDesktop.js
```diff
@@ -0,0 +1,21 @@
...
+          <ListingsHeader {...props} />
+          <ListingsList {...props} />
+        </div>
+        <div className="col-xs-6 relative overflow-y-auto b--l b--r">
+          <BodySection {...props} />
+        </div>
+      </div>
+    </div>
+  )
+}
```
justin808: This one is using `export default function(props)`. Above using `const Foobar = (props) => { }; export default Foobar`. You might want to be consistent.

You can also use  

```js
export default (props) => {

};
```

app/javascript/App/components/ExchangeListingsPage/ListingsList.js
```diff
@@ -0,0 +1,53 @@
...
+  onComponentMount() {
+    this.fetchListings();
+  }
+
+  fetchListings = () =>
+    fetch("http://localhost:3000/api/exchange_listings.json")
+      .then(response => response.json())
+      .then(result => this.onSetResult(result));
+
+  onSetResult = result =>
```
justin808: why not declare all functions

```js
onSetResult(result) {
 this.setState({
      listings: result
    });
}
```

app/javascript/App/containers/newsfeed/sagas.js
```diff
@@ -1,114 +1,122 @@
...
-  yield takeLatest('ON_FILTER_INITIALIZE', fetchCoins)
-  yield takeLatest('ON_FILTER_INITIALIZE', pollNewsItems)
-  yield takeLatest('SET_ENTITY_LIST', onSetCoinList)
-  yield takeLatest('SET_ACTIVE_ENTITY', onSetActiveCoin)
-  yield takeLatest('ON_FILTER_CHANGE', onFilterChange)
-  yield takeLatest('TOGGLE_UI', onWatchingOnly)
-  yield takeLatest('FETCH_MORE_NEWS_FEED', onScrollingToBottom)
-  yield fork(filterSagas)
-  yield fork(entitySagas)
+  yield takeLatest("ON_FILTER_INITIALIZE", fetchCoins);
```
justin808: be sure to get the project setup on linting...not sure why you switched to double quotes.

hanchang:
Ready for merge into `develop` since the feature flags are in.

app/javascript/App/bundles/ExchangeListings/components/ListingsList.jsx
```diff
@@ -0,0 +1,42 @@
...
+    super(props)
+
+    this.state = {
+      listings: props.initialListings,
+      detectedSince: null,
+    }
+  }
+
+  onComponentMount() {
+    this.fetchListings()
```
cryptoB0T: this method appears undefined. perhaps this is stubbed here for a method coming later?

app/javascript/App/containers/newsfeed/sagas.js
```diff
@@ -1,114 +1,122 @@
-import { takeLatest, select, put, fork } from 'redux-saga/effects'
```
cryptoB0T: what happened to this file, did prettier skip this one?

cryptoB0T:
@hanchang why did you rename the files to use jsx extension?

hanchang:
@alexfedoseev Would love a review when you get a chance, thanks!

app/javascript/App/bundles/ExchangeListings/client.js
```diff
@@ -0,0 +1,88 @@
...
+import ListingsList from './components/ListingsList'
+import BodySection from './components/BodySection'
+
+class ExchangeListingsPage extends Component {
+  state = { showFilterPanel: false }
+
+  toggleFilterPanel() {
+    this.setState({
+      showFilterPanel: !this.state.showFilterPanel,
+    })
```
alexfedoseev: When state depends on previous state it's highly recommended to use callback b/c by the time next state is being set `this.state` might be not what you expect (`setState` is async):

```js
this.setState(state => ({ showFilterPanel: !state.showFilterPanel }))
```

app/javascript/App/bundles/ExchangeListings/client.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  componentDidUpdate() {
+    window.addEventListener('resize', debounce(() => this.forceUpdate(), 500))
```
alexfedoseev: This is a leak: new listener is being added on every re-render and doesn't get removed. Global subscriptions should be added only once in `componentDidMount` and removed on `componentWillUnmount`.

app/javascript/App/bundles/ExchangeListings/client.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  render() {
+    const props = this.props
+
+    if (window.isMobile) {
+      console.log('mobile')
```
alexfedoseev: `console.log`

app/javascript/App/bundles/ExchangeListings/client.js
```diff
@@ -0,0 +1,88 @@
...
+      return (
+        <LayoutMobile
+          {...props}
```
alexfedoseev: I suggest to avoid destructuring of props: it's extremely hard to maintain such trees, at least w/o reliable type system (and TS is not one of them).

app/javascript/App/bundles/ExchangeListings/client.js
```diff
@@ -0,0 +1,88 @@
...
+          mainSection={
+            <Fragment>
+              <ListingsHeader
+                {...props}
+                showFilterPanel={this.state.showFilterPanel}
+                toggleFilterPanel={this.toggleFilterPanel.bind(this)}
```
alexfedoseev: Instead of using `bind` (which creates a new function on every re-render) you can use instance properties:

```js
// `this` is auto-bound
toggleFilterPanel = () => ...;

toggleFilterPanel={this.toggleFilterPanel}
```

app/javascript/App/bundles/ExchangeListings/client.js
```diff
@@ -0,0 +1,88 @@
...
+              <ListingsList {...props} />
+            </Fragment>
+          }
+          modalName="listingsModal"
+          modalSection={<BodySection {...props} mobileLayout />}
+          drawerSection={null}
+        />
+      )
+    } else if (window.isTablet) {
+      console.log('tablet')
```
alexfedoseev: `console.log`

app/javascript/App/bundles/ExchangeListings/components/ListingItem.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+
+const ListingItem = (props) => {
+  const { listing } = props
```
alexfedoseev: FYI if you want to destructure you can do this in args:

```js
const ListingItem = ({ listing }) => (
  <div />
);
```

Tho I prefer to just use `props` and `props.listing` all the way thus you don't have to worry about anything when prop gets added/removed. But it's really just personal preference.

app/javascript/App/bundles/ExchangeListings/components/Tips.jsx
```diff
@@ -0,0 +1,128 @@
...
+  width: 18,
+  paddingLeft: 6,
+  paddingTop: 2,
+  fontSize: 10,
+  verticalAlign: 'top',
+  marginTop: 2,
+}
+
+export default ({ closeTips, user, loggedIn }) => {
+  const userIsLoggedIn = !!(user || loggedIn)
```
alexfedoseev: How it can be possible that `loggedIn` is `true` but no `user` in the state (I'm not familiar w/ domain but in general it's not possible).

app/javascript/App/bundles/ExchangeListings/components/Tips.jsx
```diff
@@ -0,0 +1,128 @@
...
+import filterIcon from '../../../images/filterIcon.svg'
+import listIcon from '../../../images/listIcon.svg'
+
+const titleStyle = {
+  fontSize: '1.13rem',
+  fontWeight: 'bold',
+}
+
+titleStyle.paddingLeft = 15
+titleStyle.paddingBottom = 10
```
alexfedoseev: Why it's moved out of object definition?

app/javascript/App/bundles/ExchangeListings/components/Tips.jsx
```diff
@@ -0,0 +1,128 @@
...
+              <button
+                className="btn btn-blue btn-l w-100"
+                style={{
+                  padding: '.8rem 2rem',
+                  textTransform: 'none',
+                  fontWeight: 'bold',
+                  margin: 'auto',
+                }}
+                onClick={() => (window.location = '/sign_up')}
+              >
```
alexfedoseev: AFAICT it's just a `<a />`.

app/javascript/App/bundles/ExchangeListings/components/Tips.jsx
cryptoB0T: At one point we needed these properties for responsive design, so this was moved out.

app/javascript/App/bundles/ExchangeListings/components/Tips.jsx
cryptoB0T: You mean instead of using `button` you would use `anchor`?

app/javascript/App/bundles/ExchangeListings/client.js
ashgaliyev: @cryptoB0T event handler is not the same because it's the anonymous function. So it will register multiple times after every `componentDidUpdate`
https://jsfiddle.net/L9ahtn4k/1/
https://triangle717.wordpress.com/2015/12/14/js-avoid-duplicate-listeners/

alexfedoseev: @cryptoB0T This is a leak in few ways. MDN doc states that `If multiple identical EventListeners ... the duplicate instances are discarded`, but `debounce` produces a brand new function on every update and it makes these listeners non-identical. To make them identical this function should be created only once and attached to a class instance, i.e.:

```js
updateOnResize = debounce(() => this.forceUpdate());

componentDidUpdate() {
  window.addEventListener('resize', this.updateOnResize);
}

// now `this.updateOnResize === this.updateOnResize` is always `true`
```

But even then there still is a leak: when component is unmounted from the DOM all added listeners are still around and stuck here forever since all instances required for removal are not available anymore and unmounting of a node from the DOM doesn't trigger auto-removal of all attached listeners.

So, instead of performing wasteful ops in `componentDidUpdate`, it's better to add one single listener in `componentDidMount` then on every `resize` this debounced handler will be triggered, plus to remove it in `componentWillUnmount`.

PR #277, opened 2018-08-03
Create rake task to ingest coinmarketcal data.

anthotsang:
https://app.asana.com/0/709175359523131/742654163136031


anthotsang:
`2018-08-04T00:18:54.468780+00:00 app[scheduler.1372]: PG::UndefinedFunction: ERROR:  function dblink(unknown, unknown) does not exist`

Should have been added by a db migration. Should I just try to call it as an sql command at the start of the rake task instead?

hanchang:
@anthotsang Stack Overflow is showing this answer:

`The problem was that dblink was not created in public schema as expected. so have to qualify with someschema.dblink(...)`

https://stackoverflow.com/questions/50936251/error-function-dblinkunknown-unknown-does-not-exist

(It's the last comment, hidden by the threading.)

Maybe give that as shot?

.env.example
```diff
@@ -27,6 +27,11 @@ MAILGUN_SMTP_SERVER=
 NEWS_TAGGER_ENDPOINT=
 NEWS_TAGGER_BASIC_AUTH_USERNAME=
 NEWS_TAGGER_BASIC_AUTH_PASSWORD=
+RDS_NAME=
```
hanchang: Let's change the ENV var names to CALENDAR_DB_XXX since we'll almost assuredly have multiple RDS databases connected in the near future.

db/migrate/20180803130417_add_column_to_calendar_events_table.rb
```diff
@@ -0,0 +1,5 @@
+class AddColumnToCalendarEventsTable < ActiveRecord::Migration[5.1]
+  def change
+    add_column :calendar_events, :import_id, :bigint
```
hanchang: We should probably add an index on `import_id` otherwise query performance can deteriorate when the table starts getting large especially since we're going to be sorting by `import_id`.

lib/tasks/coinmarketcal.rake
```diff
@@ -0,0 +1,39 @@
...
+
+    user = User.find_by_email("coinmarketcal@coinfi.com")
+    unless user
+      user = User.create(
+        email: "coinmarketcal@coinfi.com",
+        password: Devise.friendly_token[0, 20]
+      )
+    end
+
+    @connection.execute("CREATE OR REPLACE VIEW coinmarketcal_event_view AS
```
hanchang: LOVE LOVE LOVE the usage of a View! 😻

anthotsang:
@hanchang 
Seems there's only a public schema.
Looks like none of the db migrations ran (i.e., import_id also missing).
I came across this https://github.com/mbleigh/acts-as-taggable-on/issues/866
but we're already using acts-as-taggable-on 5.0

PR #282, opened 2018-08-08
fix for bug where users are unable to scroll news in firefox

cryptoB0T:


app/javascript/App/components/NewsfeedPage/NewsList.js
```diff
@@ -20,6 +20,7 @@ class NewsList extends Component {
       this.setState({ initialRender: false })
     }, 60000)
     this.mountOnScrollHandler()
+    document.querySelector('#newsfeed').style.height = `${window.outerHeight}px`
```
hanchang: @cryptoB0T Shouldn't this technically be `window.innerHeight` minus the height of the topnav minus the height of the searchbar? Otherwise the scrollbar behaves oddly.

cryptoB0T: @hanchang nice catch, thanks!

app/javascript/App/components/NewsfeedPage/NewsList.js
```diff
@@ -20,6 +20,8 @@ class NewsList extends Component {
       this.setState({ initialRender: false })
     }, 60000)
     this.mountOnScrollHandler()
+    const newsfeedElem = document.querySelector('#newsfeed')
+    newsfeedElem.style.height = `${newsfeedElem.offsetHeight}px`
```
hanchang: @cryptoB0T This still doesn't work for me... the scrollbar goes past the screen and users don't see the spinning loading icon when loading the infinite scroll.

cryptoB0T: @hanchang I see what you mean. as the user scrolls down, they page loads more items and the height of the container increases. let me think about this, I've got to get this to work..

app/javascript/App/components/NewsfeedPage/NewsList.js
```diff
@@ -21,7 +21,7 @@ class NewsList extends Component {
     }, 60000)
     this.mountOnScrollHandler()
     const newsfeedElem = document.querySelector('#newsfeed')
-    newsfeedElem.style.height = `${newsfeedElem.offsetHeight}px`
+    newsfeedElem.style.maxHeight = `${newsfeedElem.offsetHeight}px`
```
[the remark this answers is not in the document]
hanchang: Yup that's better.

PR #289, opened 2018-08-10
fix selecting feed sources in filter panel

cryptoB0T:


app/javascript/App/components/ItemSelectorAlt.js
```diff
@@ -44,10 +45,14 @@ export default class ItemSelectorAlt extends Component {
       )
     } else {
       return (
-        <a className="mid-gray" onClick={() => this.add(item)}>
-          <input type="checkbox" className="mr2 w-auto" />
+        <label
```
hanchang: This is good, but the same `<label>` wrap needs to be applied if the item is selected.

cryptoB0T: nice find!

PR #293, opened 2018-08-13
Feature/update top nav/#772319848111012

cryptoB0T:


app/assets/stylesheets/modules/mobile-nav.scss
```diff
@@ -15,6 +15,12 @@
   z-index: 98;
   transition: all 0.5s;
   opacity: 0;
+  overflow-y: scroll;
+  > div {
```
hanchang: What is this padding for?

app/views/components/_main_nav.slim
```diff
@@ -8,12 +8,14 @@ nav class=("#{defined?(klass) ? klass : ''}")
       li
         = nav_link 'News', '/news'
     li
-      a href="https://blog.coinfi.com" target="_blank" Blog
+      = nav_link 'Events', '/events'
```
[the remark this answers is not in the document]
cryptoB0T: @hanchang if we put in the feature flag checks the rails app will crash because it's undefined. do you want it in there commented out?

app/assets/stylesheets/modules/mobile-nav.scss
cryptoB0T: without this padding it won't show the coinfi logo on shorter heights, like mobile landscape

PR #294, opened 2018-08-13
show read/unread news items

cryptoB0T:


app/javascript/App/components/NewsfeedPage/NewsListItem.js
```diff
@@ -26,12 +27,53 @@ const NewsListItem = (props) => {
...
+
   return (
     <div className={className} style={{ height: props.height || 'auto' }}>
       <div className="pa-default">
         <div
           className="pointer"
           onClick={() => {
+            const newsId = newsItem.get('id')
+            const readNewsData =
+              JSON.parse(localStorage.getItem('readNews')) || []
```
hanchang: Maybe leverage something like https://github.com/pamelafox/lscache instead of DIY?

cryptoB0T: I like the recommendation! this lib looks pretty cool. i'm not sure how useful the mem-cache-like api will be for this particular use-case. the disadvantage is we can store less items because we store key, value, and time. if we omit storing time, then we lose the ability to expire the oldest items. my concern would be potentially degrading performance as we fill up the local storage faster than just storing the raw index.

PR #341, opened 2018-08-24
[WIP] clean up code using react context

cryptoB0T:
* reduce amount of code
* reduce chances of bugs
* avoid passing around props

### DIFF
--- a/app/javascript/App/bundles/ExchangeListings/client.js
+++ b/app/javascript/App/bundles/ExchangeListings/client.js
@@ -80,6 +80,7 @@ class ExchangeListingsPage extends Component {
 
   changeSymbol = (data) => {
     const selectedSymbols = data.map((item) => item.value)
+
     this.setState({
       selectedSymbols: selectedSymbols,
     })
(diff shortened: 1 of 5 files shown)


app/javascript/App/bundles/ExchangeListings/client.js
```diff
@@ -7,6 +7,7 @@ import ListingsHeader from './components/ListingsHeader'
 import ListingsList from './components/ListingsList'
 import BodySection from './components/BodySection'
 import localAPI from '~/lib/localAPI'
+import { Provider } from '~/bundles/common/context'
```
cryptoB0T: we could give this assignment a long descriptive name like `ExchangeListingsContextProvider` but i don't think that's helpful, given that the context will only be visible to components descending from `ExchangeListings/client`. 

when we add `CoinListContextProvider` we can use that name. In subsequent components we could have explicit names for descending component trees while the component in question is the default provider.

app/javascript/App/bundles/ExchangeListings/client.js
```diff
@@ -153,27 +154,30 @@ class ExchangeListingsPage extends Component {
...
-                changeExchange={this.changeExchange}
-                filterDates={this.filterDates}
-                selectedItems={selectedItems}
-                selectedSymbols={this.state.selectedSymbols}
-                selectedExchanges={this.state.selectedExchanges}
-                exchangeSlugs={this.state.exchangeSlugs}
-                resetFilters={this.resetFilters}
-              />
+              <Provider value={context}>
+                <ListingsHeader showFilterPanel={this.state.showFilterPanel} />
```
cryptoB0T: this property is the only one that is used at that level. the other 12 were just being passed through.

hanchang: Nice, great cleanup! ✨

app/javascript/App/bundles/ExchangeListings/client.js
hanchang: I agree with your statement except for the fact that the file is located in `bundles/common/`.

As a result, from my understanding it's impossible to know that `context.js` is associated with ExchangeListings without deeply diving into the code...

alexfedoseev: IMO `ExchangeListingsContextProvider` is a very good name. It's very easy to reason about a code where it's used. But in current implementation it's quite hard to understand what this component is even by looking at its source. And it's very easy, later on, to get into confusing situations like when new junior developer will see `bundles/common/context` module and decide that it can be re-used for anything (e.g. render multiple different providers of different data or consume whatever he/she needs at the moment) b/c it looks like "I'm a generic thing, re-use me!".

All in all, I'd use `ExchangeListingsContextProvider` and most likely moved it to `bundles/ExchangeListings` b/c it looks like it's coupled to this bundle (but I might be wrong here, judging by names).

alexfedoseev: In case if long name confuses you, the fact: the whole Facebook's frontend codebase (it's monorepo that holds all hack/react/react-native/reason/etc code of facebook.com/messanger.com) consists of modules with uniq names: every single component has its own descriptive name and it's enforced by the tooling.

PR #370, opened 2018-09-01
[DNM] Remove npm from package.json in favor of yarn

hanchang:


hanchang:
Combined with https://github.com/coinfi/coinfi/pull/371

PR #376, opened 2018-09-04
Feature/refactor coinsearch

ashgaliyev:


app/assets/stylesheets/components/autosuggest.scss
```diff
@@ -0,0 +1,57 @@
+.autosuggest {
```
hanchang: We may need to rename this in the future if there's other uses for the `react-autosuggest` since this style is probably more geared towards displaying the Coin search results, but for now it's fine.

hanchang:
🏆 Nice!

PR #380, opened 2018-09-05
Add CoinList to ExchangeListings

ashgaliyev:


app/javascript/App/bundles/common/components/CoinListItem.js
```diff
@@ -1,12 +1,14 @@
...
 import PercentageChange from './PercentageChange'
 import WatchButton from './WatchButton'
 
 export default (props) => {
   const { coin, user, onClick } = props
   let klass = 'pa-default b--b flex items-center pointer'
-  //if (isActiveEntity({ type: 'coin', id: coin.get('id') })) klass += ' bg-foam'
+
+  if (props.isSelected) {
+    klass += ' bg-foam'
```
hanchang: Is this something that `classNames` can help with?

ashgaliyev: Yes, it definitely needs to be rewritten with `classnames` function to make it more readable. Thanks!

PR #386, opened 2018-09-09
News Refactoring UI Improvements

ashgaliyev:


app/javascript/App/bundles/NewsfeedPage/NewsListHeader.tsx
```diff
@@ -34,41 +37,20 @@ export default class NewsListHeader extends React.Component<IProps, {}> {
     return (
       <>
         {window.isMobile && (
-          <SectionHeaderTight>
```
hanchang: Thanks for getting rid of this @ashgaliyev !

hanchang:
This LGTM now - the build failure doesn't seem to be accurate so I think we can safely ignore.

@jessehon any thoughts?

PR #391, opened 2018-09-10
Consolidate news categories #798338880309897

jessehon:
https://app.asana.com/0/683768070860536/798338880309897/f

Before migration
image

After migration
image


lib/tasks/data_migrations/04_consolidate_news_categories.rake
```diff
@@ -0,0 +1,52 @@
...
+        'Forks',
+        'Adoption',
+        'Token Supply Changes',
+      ]
+      project_announcements_categorizations = NewsItemCategorization
+        .joins(:news_category)
+        .where(news_categories: {name: PROJECT_ANNOUNCEMENTS_CATEGORIES})
+
+      update_count = 0
+      project_announcements_categorizations.each do |categorization|
```
hanchang: We should probably switch to `find_each` here to conserve memory.

lib/tasks/data_migrations/04_consolidate_news_categories.rake
```diff
@@ -0,0 +1,52 @@
...
+        categorization.update!(news_category: project_announcements_category)
+        update_count += 1
+      end
+      puts "Updated #{update_count} NewsItemCategorization records"
+
+      # Remove replaced categories
+      replaced_categories = PROJECT_ANNOUNCEMENTS_CATEGORIES
+      destroy_count = 0
+      NewsCategory.where(name: replaced_categories).each do |category|
```
hanchang: This is probably fine since there aren't that many `NewsCategories` anyways.

PR #394, opened 2018-09-10
Remove `main` top margin

hanchang:
Before:
lncy2ptw

After:
screen shot 2018-09-10 at 11 39 18 am


jessehon:
The margin was added and was supposed to be applied for mobile screens only as the nav will transition to `fixed` positioning

hanchang:
@jessehon Got it, thanks for the context! Rejiggered to meet both requirements, lemme know if you see anything else amiss.

PR #400, opened 2018-09-11
Basic integration tests #811191988078851

jessehon:
https://app.asana.com/0/683768070860536/811191988078851

### DIFF
--- a/config/initializers/launch_darkly.rb
+++ b/config/initializers/launch_darkly.rb
@@ -1,3 +1,3 @@
 # This is actually handled on production by config/puma.rb in the `on_worker_boot` section.
 # This configuration is just for localhost development.
-$launch_darkly = LaunchDarkly::LDClient.new(ENV.fetch("LAUNCHDARKLY_SDK_KEY")) if Rails.env.development?
+$launch_darkly = LaunchDarkly::LDClient.new(ENV.fetch("LAUNCHDARKLY_SDK_KEY")) if Rails.env.test? || Rails.env.development?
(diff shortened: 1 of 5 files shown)


config/initializers/launch_darkly.rb
```diff
@@ -1,3 +1,3 @@
 # This is actually handled on production by config/puma.rb in the `on_worker_boot` section.
 # This configuration is just for localhost development.
-$launch_darkly = LaunchDarkly::LDClient.new(ENV.fetch("LAUNCHDARKLY_SDK_KEY")) if Rails.env.development?
+$launch_darkly = LaunchDarkly::LDClient.new(ENV.fetch("LAUNCHDARKLY_SDK_KEY")) if Rails.env.test? || Rails.env.development?
```
hanchang: Maybe it's better just to do `unless Rails.env.production?` and delete the comment on line 2?

hanchang: Actually I'm retarded - I think this needs to happen always, since if we run rake tasks on prod LaunchDarkly won't be loaded correctly.

So the correct thing is to just remove the `if` statement and always load `$launch_darkly` both here in the initializer and in the Puma config.

hanchang: The reason we need to also load it in the Puma config is because Puma is multithreaded so unless the instantiation happens there, sometimes a Puma thread won't get the `$launch_darkly` variable initialized :(

jessehon: Yea okay, makes sense. Maybe worth exploring a better solution at a later stage though so we don't need to duplicate this logic

PR #402, opened 2018-09-11
Profile link in nav #778628919383293

jessehon:
https://app.asana.com/0/683768070860536/778628919383293

Desktop
image

Mobile
image


app/javascript/App/bundles/common/components/NavUser.tsx
```diff
@@ -0,0 +1,156 @@
...
+  ListSubheader,
+} from '@material-ui/core'
+import {
+  AccountCircle,
+  KeyboardArrowDown,
+  KeyboardArrowUp,
+} from '@material-ui/icons'
+import { MenuProps } from '@material-ui/core/Menu'
+import { createStyles, withStyles, withTheme } from '@material-ui/core'
+import { PROFILE_EDIT_URL, LOGOUT_URL } from '~/constants'
```
hanchang: Love this! 😻

app/javascript/App/index.js
```diff
@@ -61,6 +65,7 @@ const injectComponents = () => {
       if (withStore) {
         if (false) {
           // ReactDOM.render(
+          //   <MuiThemeProvider theme={theme}>
```
hanchang: Thanks for making sure to add this to the new refactor version :bow:

app/assets/stylesheets/components/navigation.scss
```diff
@@ -55,6 +55,33 @@ $nav-menu-more-width: 150px;
...
 
+.nav-menu-actions {
+  position: absolute;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  padding: 1rem;
+
+  &__btn {
+    @extend .btn, .btn.btn-sm;
```
hanchang: Getting this warning:

```
DEPRECATION WARNING on line 66 of /Users/halo/Code/coinfi/coinfi/app/assets/stylesheets/components/navigation.scss:
Extending a compound selector, .btn.btn-sm, is deprecated and will not be supported in a future release.
Consider "@extend .btn, .btn-sm" instead.
```

app/javascript/App/bundles/common/components/NavUser.tsx
```diff
@@ -0,0 +1,156 @@
...
+  )
+
+  return (
+    <div className={classes.root}>
+      <Avatar className={classes.avatar}>
+        <AccountCircle />
+      </Avatar>
+
+      {menuOpen ? (
+        <IconButton
```
hanchang: Would be nice to style the `IconButton` to be a bit less wide, since the spacing looks a bit weird right now (i.e. the chevron is kinda far from the user profile icon) - not a huge deal though!

hanchang:
Actually @jessehon the build is failing due to OOM issues so we may need to address that:
https://github.com/rails/webpacker/issues/1284

jessehon:
Found the cause for this OOM issue after running the `webpack-bundle-analyzer` plugin. Instead of importing specific components, I imported the entire `@material-ui/core` and `@material-ui/icons` libraries, which turned out to be around 5mb. 

You can see that several other external libraries are also quite large, which we may consider performing similar improvements to e.g. lodash

Bundle analysis
image

Bad imports
image

hanchang:
Nice, great find! 💪 Build seems to be working so we should be good to go here.

PR #406, opened 2018-09-13
Tag coin tweets as project announcements #817501498201006

jessehon:
https://app.asana.com/0/683768070860536/817501498201006


app/services/assign_news_item_default_news_categories.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def call
+    if is_coin_project_tweet?
+      project_announcements_category = NewsCategory.find_project_announcements!
+
+      # Assign if does not already exist
+      unless @news_item.news_categories.exists?(project_announcements_category.id)
+        @news_item.news_categories << project_announcements_category
+        @news_item.last_machine_tagged_on = Time.now
```
jessehon: @hanchang I just realized, this field is used for tagging with coin mentions and not categories right?

hanchang: @jessehon Yes, great catch - machine tagging only handles tagging Coins, there's no capability for machine tagging NewsCategories at the moment.

hanchang: So we should not mark NewsItems as `last_machine_tagged_on` since it's assumed that all "Project Announcements" NewsCategories will be automatically marked.

hanchang: Instead, the `last_machine_tagged_on` field should only be set by the NewsTagger application.

PR #408, opened 2018-09-13
Feature/migrate to react on rails

ashgaliyev:


Gemfile
```diff
@@ -34,6 +34,8 @@ gem 'redis-rails' # Remove when upgrading to Rails 5.2 since it is built-in
 # Use Capistrano for deployment
 # gem 'capistrano-rails', group: :development
 
+gem 'react_on_rails', '11.1.3'
```
justin808: 11.1.4 now.

app/assets/stylesheets/base/general.scss
```diff
@@ -1,3 +1,7 @@
+script {
+  display: none !important;
+}
+
```
ashgaliyev: There is the rule in css on some pages:
```
... > * {
  display: flex
}
```
it caused displaying script tag with all contents.

app/javascript/App/bundles/common/components/WatchButton.tsx
```diff
@@ -71,4 +72,6 @@ const WatchButton = ({ coin, hasText, loggedIn }: Props) => {
   )
 }
 
+ReactOnRails.register({ WatchButton })
```
ashgaliyev: It makes the component to be available for `react_component` function in rails

app/javascript/App/components/CalendarPage/index.js
```diff
@@ -143,4 +145,6 @@ class CalendarPage extends Component {
   }
 }
 
-export default calendarContainer(CalendarPage)
+const CalendarPageWithStore = withStore(calendarContainer(CalendarPage))
+
+ReactOnRails.register({ CalendarPage: CalendarPageWithStore })
```
ashgaliyev: This component uses redux. So, I'm wrapping it into `withStore` HOC to make it works until it will be refactored into `non-redux` version.

Also, the exporting of `CalendarPage` is not necessary because this component is used only by `react_component` function.

app/javascript/App/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+import theme from '~/theme'
+
+import ExchangeListingsPage from '~/bundles/ExchangeListings/ExchangeListingsCointainer'
+import NewsfeedPageNew from '~/bundles/NewsfeedPage/NewsfeedPageContainer'
+import CoinIndex from './components/CoinIndex'
+import CoinListContainer from '~/bundles/common/containers/CoinListContainer'
+import NewsfeedContainer from '~/bundles/NewsfeedPage/NewsfeedContainer'
+import ReactOnRails from 'react-on-rails'
+
+const App = (props, railsContext) => (
```
ashgaliyev: This file now contains only main app with react router. All other components are imported in `startup` file to make the structure cleaner.

app/javascript/App/startup.js
```diff
@@ -0,0 +1,10 @@
+// react-router app
+import './index'
+
+// non-react-router components
+import './bundles/common/containers/NavUserContainer'
+import './components/CoinShow'
+import './components/IcoFilters'
+import './bundles/common/containers/FlashMessageListContainer'
+import './bundles/common/components/WatchButton'
+import './components/CalendarPage'
```
ashgaliyev: Now, this is the entry point for webpack. If some new separate component will be created it must be imported here in `non-react-router components`.

app/javascript/App/withStore.tsx
```diff
@@ -0,0 +1,50 @@
...
+              </PersistGate>
+            </Provider>
+          </CoinListContainer>
+        </MuiThemeProvider>
+      )
+    }
+  }
+}
+
+export default withStore
```
ashgaliyev: Here is the `store` wrapper for redux-dependent components. It's like in the old `App/index.js` but now it's a HOC

app/javascript/packs/application.js
```diff
@@ -1,7 +1,27 @@
+import 'babel-polyfill'
```
ashgaliyev: I've moved it here from `App/index.js` because it's not related to app business logic.

app/javascript/packs/application.js
```diff
@@ -1,7 +1,27 @@
...
+  window.isTablet = window.screenSize === 'm'
+  window.isDesktop = window.screenSize === 'l'
+}
+
+document.addEventListener('DOMContentLoaded', () => {
+  setScreenSize()
+})
+
+window.addEventListener('resize', debounce(setScreenSize, 400))
+window.addEventListener('resize', debounce(scrollHelper, 400))
```
ashgaliyev: I was thinking about to add it into some react component. But it's not related to react app. So, I just moved it here from old `App/index.js`.

Gemfile
ashgaliyev: Done

app/assets/stylesheets/base/general.scss
jessehon: Thanks for the clarification. Yeah, not sure why such a open css rule was initially added but we'll definitely try and remove it in a later refactor. The `display: none !important` fix is fine as a workaround for now.

app/javascript/App/bundles/common/components/WatchButton.tsx
jessehon: Is it better to move these `ReactOnRails.register` calls into a single location e.g. the webpack entry point `startup.js`

app/javascript/App/components/CalendarPage/index.js
jessehon: I think it is still better to export just to maintain consistency with all the other components

app/javascript/App/index.tsx
jessehon: Thats much cleaner, thanks

PR #409, opened 2018-09-14
Consistent social feeds display #788048874942004

jessehon:
https://app.asana.com/0/683768070860536/788048874942004


app/javascript/App/bundles/NewsfeedPage/NewsfeedPage.tsx
```diff
@@ -191,18 +209,25 @@ class NewsfeedPage extends React.Component<Props, State> {
   }
 
   public componentDidUpdate(prevProps, prevState, snapshot) {
+    // Check if `coinSlug` in the route changed
```
hanchang: Thanks for the explainer comments, super helpful ❤️

app/javascript/App/bundles/NewsfeedPage/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const getDefaultFilters = (): Filters => ({
+  categories: [],
+  coinSlugs: [],
+  feedSources: [],
+  publishedSince: null,
+  publishedUntil: null,
+})
+
+export const TOP_COINS = [
```
hanchang: Hm, the issue with this implementation is that because the crypto markets are so volatile, even the top 5 coins by market cap will change so it's dangerous to hard code.

How difficult would it be to make this a dynamic check instead?

hanchang: Maybe just have a query to grab and cache the top 5 market cap coins daily.

hanchang: If this implementation change will take longer than 2 hours let's just spin up a new ticket for it on the next sprint.

jessehon: To save time I will implement this a root prop and simply pass it via prop drilling into the required components. We can clean this up together with the rest of the NewsfeedPage when we do our refactor.

PR #410, opened 2018-09-14
Flag to toggle human machine tagged news coin mentions #728802939714043

jessehon:
https://app.asana.com/0/683768070860536/728802939714043

Added new required environment variable NEWS_COIN_MENTION_TAG_SCOPE with values "human", "machine" and "combo"


app/models/news_item.rb
```diff
@@ -43,8 +47,29 @@ def self.chart_data(is_topcoin = false)
       .order_by_published(:asc)
   end
 
-  def coin_link_data
-    coins.map { |coin| coin.as_json(only: [:symbol, :slug, :id] ) }
+  def tag_scoped_coins
+    case ENV['NEWS_COIN_MENTION_TAG_SCOPE']
```
hanchang: We try to use `ENV.fetch('NEWS_COIN_MENTION_TAG_SCOPE')` when the env var is required so that the app throws an error if it's missing.

app/controllers/api/news_items_controller.rb
```diff
@@ -67,9 +67,23 @@ def index
   private
 
   def serialized(obj)
-    obj.as_json(
+    data = obj.as_json(
       only: %i[id title summary feed_item_published_at updated_at url content],
-      methods: %i[coin_link_data categories]
+      methods: %i[tag_scoped_coin_link_data categories]
     )
+    format_item = Proc.new do |item, *args||
```
hanchang: Is there supposed to be two bars? `|item, *args||`

PR #412, opened 2018-09-17
Update get screen size from window to context #823462745446780

jessehon:
https://app.asana.com/0/683768070860536/823462745446780


app/javascript/App/components/ItemSelectorCategory.js
```diff
@@ -32,50 +36,51 @@ export default class ItemSelectorCategory extends Component {
...
-                  <ItemButton item={itemLabel} style={{cursor:'pointer'}} />
-                </div>
-              )
-            })}
-          </div>);
+          return (
+            <div className="row" key={index}>
+              {itemGroups.map((item, innerIndex) => {
+                let itemLabel
+                if (item === 'Events - Conferences, Meetups, Launches, etc.') {
```
hanchang: I think we can safely nuke these with the release of the new NewsCategories.

app/javascript/App/components/CalendarPage/CalendarList.js
```diff
@@ -27,7 +27,7 @@ class CalendarList extends Component {
```
hanchang: Does this file also need `import withDevice` and a corresponding HOC wrapper in the `export default`?

hanchang:
Looks like a great start!

app/javascript/App/components/CalendarPage/CalendarList.js
jessehon: Yes it does, missed that change

app/javascript/App/components/ItemSelectorCategory.js
jessehon: Still referenced in the `FilterPanel` by the `CalendarPage` it seems

hanchang: Ah, thanks for the catch - we can safely ignore Calendar as it basically needs to be rebuilt sans Redux anyways.

PR #421, opened 2018-09-25
Bugfix - Fix News Filters

hanchang:
https://app.asana.com/0/683768070860536/836042970925458


Procfile.dev
```diff
@@ -1,5 +1,5 @@
 # You can run these commands in separate shells
-web: rails s -p 3000
+web: rails s webrick -p 3000
```
jessehon: Can you clarify what the reason was for changing this to `webrick`

PR #423, opened 2018-09-25
Add `rel="canonical"` to news articles #828528854649563

hanchang:
https://app.asana.com/0/683768070860536/828528854649563


app/controllers/news_controller.rb
```diff
@@ -31,9 +29,10 @@ def show
         .order_by_published
         .limit(25)
     )
-    @news_item_data = serialize_news_items(
-      NewsItem.published.find(params[:id])
-    )
+    @news_item = NewsItem.published.find(params[:id])
```
jessehon: probably doesn't need to be an instance variable if not used outside the method

hanchang: Ah nice thanks for the catch!

PR #427, opened 2018-09-26
Redirect coin show by id to slug #830067084885308

jessehon:
https://app.asana.com/0/683768070860536/830067084885308

### DIFF
--- a/app/controllers/coins_controller.rb
+++ b/app/controllers/coins_controller.rb
@@ -1,4 +1,6 @@
 class CoinsController < ApplicationController
+  before_action :set_coin, only: [:show]
+
   def index
     @coin_count = Coin.legit.listed.count
     @coins = serialize_coins(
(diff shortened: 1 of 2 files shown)


app/controllers/coins_controller.rb
```diff
@@ -56,6 +57,23 @@ def show
...
+    if coin_by_slug
+      @coin = coin_by_slug
+      return
+    end
+
+    # If we don't find matches for slug, we can safely assume it is an id
+    coin_id = coin_id_or_slug
+    coin_by_id = Coin.find(coin_id)
+    # 301 redirect to the same action with the coin slug for SEO purposes
+    redirect_to action: action_name, id_or_slug: coin_by_id.slug, status: :moved_permanently
```
jessehon: Sweet, yeah I wanted to double check with you about that before I made it a permanent redirect

hanchang: Much appreciated, confirmed we want a 301

PR #445, opened 2018-10-02
Cleanup Refactoring Style Consistency

anthotsang:
https://app.asana.com/0/683768070860536/817501498200989


app/javascript/App/bundles/common/utils/images.ts
```diff
@@ -0,0 +1,91 @@
+export const bellIcon = `data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" width="14" height="17" viewBox="0 0 14 17">
```
jessehon: I think it may be better to keep these as separate `.svg` files than merged into a single `.ts`. I could see it easily getting out of hand. It's probably not too bad to keep them as `require(...)` calls now until we configure webpack to accept importing `.svg`s

package.json
```diff
@@ -84,6 +79,14 @@
...
     "@storybook/addons": "^3.4.8",
     "@storybook/react": "^3.4.8",
+    "@types/axios": "^0.14.0",
+    "@types/lodash": "^4.14.116",
+    "@types/qs": "^6.5.1",
+    "@types/react": "^16.4.9",
+    "@types/react-currency-formatter": "^1.1.0",
+    "@types/react-dom": "^16.0.7",
+    "@types/react-router-dom": "^4.3.0",
+    "@types/react-select": "^2.0.2",
```
jessehon: I think we moved these into `dependencies` because we were having issues with our Heroku production builds

app/javascript/App/bundles/common/components/LoadingIndicator.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react'
+
+const loadingImg = `<svg version="1.1" id="loader-1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
```
jessehon: I think it may be better to keep this as a separate `.svg` file that is imported in

package.json
jessehon: Relevant issue https://github.com/rails/webpacker/issues/117

anthotsang: ah i see. i'll move them back. that might be why the build is failing

app/javascript/App/bundles/common/utils/images.ts
anthotsang: I could separate them to a file a piece rather than all in one? Not sure if it's still an issue with server side rendering implemented, but before the images would sometimes pop in since they were referencing an actual image rather than inline.

jessehon: I think there is a webpack plugin/loader for us to do that for all `.svg`s automatically. Maybe beyond the scope of this PR though so I would keep these as `.svg`s and deal with it later. What do you think @hanchang?

hanchang: Yes, def beyond scope of this PR/ticket, let's keep as is for now and @anthotsang please spin up a new ticket to handle this in the Platform board under the Inbox row and we'll prioritize for another sprint, thanks!

anthotsang: ok restored to .svg files and created ticket for the future.

app/javascript/App/withLegacyCombinedProviders.tsx
```diff
@@ -2,21 +2,15 @@ import * as React from 'react'
...
-    const TargetComponentWithRedux = withReduxStore(
-      withLegacyAppReduxContainer(TargetComponent),
-    )
-
     return (
       <DeviceProvider {...railsContext.deviceProviderProps}>
         <MuiThemeProvider theme={theme}>
           <CoinListContainer loggedIn={!!props.user}>
-            <TargetComponentWithRedux {...props} />
+            <TargetComponent {...props} />
```
hanchang: Do we need this file anymore either?

app/javascript/App/bundles/common/normalizers/index.js
```diff
@@ -0,0 +1,5 @@
+import coins from './coins'
+
```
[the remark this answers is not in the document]
anthotsang: It used to have more normalizers but they were all removed. Will remove it and directly access the coin normalizer.

app/javascript/App/withLegacyCombinedProviders.tsx
anthotsang: Not entirely sure. When I did the refactor there was still a withCombinedProviders but it looks like @jessehon got rid of it in #444 
Will look into it

anthotsang: Looks Like CalendarPage will lose CoinListContainer, which is provided by withLegacyCombinedProviders but not withClientProviders but might as well remove it and figure out how to reconnect it later.

hanchang:
This looks good to me but since it's such a massive change we need to do a bunch of QA to make sure everything works still.

I'm also aware of the importance of merging this in quickly to prevent such a large long-standing branch so let's aim to get this in by end of week.

I'll try to get the automated tests in place by tomorrow so we can have more confidence when merging this as well.

PR #449, opened 2018-10-03
Set up Rollbar to catch client side Javascript errors

anthotsang:
https://app.asana.com/0/683768070860536/844582750600046

Some static pages, like the landing page, don't have rollbar, if that's okay.


app/views/layouts/_tracking_scripts.html.erb
```diff
@@ -1,18 +1,4 @@
...
-// End Rollbar Snippet
-</script>
-
```
hanchang: Ah nice, leveraging the Ruby gem's ability to integrate with JavaScript 💪 
https://docs.rollbar.com/docs/integration-with-rollbarjs

config/initializers/rollbar.rb
```diff
@@ -59,7 +59,9 @@
   config.js_options = {
     accessToken: ENV.fetch('ROLLBAR_CLIENT_ACCESS_TOKEN'),
     captureUncaught: true,
-    hostWhiteList: ['staging.coinfi.com', 'www.coinfi.com', 'sale.coinfi.com'],
+    captureUnhandledRejections: true,
+    logLevel: 'error',
+    hostWhiteList: ['coinfi.com', 'herokuapp.com'],
```
hanchang: Does an entry in `hostWhiteList` include all subdomains?

If not, it would be useful to be able to test this on the develop and staging subdomains as well, and the herokuapp.com domain would be useless unfortunately :(

anthotsang: Yeah according to the [documentation](https://docs.rollbar.com/docs/reduce-noisy-javascript-errors#section-whitelist-specific-domains) it treats it like regex:

>When hostWhiteList has been configured, Rollbar will enumerate over the stack frames in each error. At least one of the frames must contain a filename that contains at least one of the strings in this configuration option. The items in the array are compiled into a regex that is used to compare against the filenames.

hanchang: Awesomesauce, thanks!

PR #450, opened 2018-10-03
Fix signals client side styles #848497549807770

jessehon:
https://app.asana.com/0/683768070860536/848497549807770


anthotsang:
Still seems to occur:
>Warning: Prop `className` did not match. Server: "SignalPopoverText-text-2" Client: "SignalPopoverText-text-35"

Anything I could do to help you debug/figure out the root cause?

jessehon:
> Still seems to occur:
> 
> > Warning: Prop `className` did not match. Server: "SignalPopoverText-text-2" Client: "SignalPopoverText-text-35"
> 
> Anything I could do to help you debug/figure out the root cause?

Not sure how to address that unfortunately so open to ideas if you have any. 

The `JssProvider` change here should prevent any clashes within client rendered styles though so hopefully worse case that means the component tree will re-render when react hydrates but at least the UI wouldn't break.

jessehon:
The client rendered styles might still clash with the server rendered styles though. The suggested approach seems to be removing the server rendered styles when the client components mount. However since we hav multiple root components rendered on a page, not sure if there will be a race between components mounting their client styles and removing server styles.

PR #452, opened 2018-10-03
Link Coin Tags on the News Content pane back to the Coin News (news/bitcoin) pages

anthotsang:
https://app.asana.com/0/683768070860536/830067084883310

### DIFF
--- a/app/javascript/App/bundles/NewsfeedPage/NewsfeedPage.tsx
+++ b/app/javascript/App/bundles/NewsfeedPage/NewsfeedPage.tsx
@@ -16,7 +16,8 @@ import withDevice from '~/bundles/common/utils/withDevice'
 import EventListener from 'react-event-listener'
 
 import { NewsItem, ContentType, Filters } from './types'
-import { CoinList, CoinWithDetails } from '../common/types'
+import { CoinWithDetails } from '../common/types'
+import { CoinLinkData } from '~/bundles/common/types'
 import {
   getDefaultFilters,
   mergeInitialSocialSourcesForCoinsFilter,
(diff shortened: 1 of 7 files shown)


app/javascript/App/bundles/NewsfeedPage/NewsfeedPage.tsx
```diff
@@ -317,6 +318,10 @@ class NewsfeedPage extends React.Component<Props, State> {
 
   public render() {
     if (this.props.isMobile) {
+      const coinClickHandler = (coinData: CoinLinkData) => {
+        this.props.history.push(`/news/${coinData.slug}`)
+        this.setState({ ActiveMobileWindow: 'None' })
```
hanchang: @jessehon Is this necessary for all links? Not sure exactly when/how/where to use this...

jessehon: At the moment, I believe we still need `ActiveMobileWindow` otherwise the wrong pane would end up showing. Ideally this should just be inferred from the router state but for now, we still need to manually update it.

hanchang: Awesome, thanks for the context! 🙏

app/javascript/App/bundles/common/components/CoinTags.tsx
```diff
@@ -1,21 +1,36 @@
...
   <div>
-    {itemWithCoinLinkData.coin_link_data.map((data, index) => (
-      <a
-        key={index}
-        className="tag pointer" /* onClick={() => selectCoin(data)}*/
-      >
-        {data.symbol}
-      </a>
-    ))}
+    {itemWithCoinLinkData.coin_link_data.map((data, index) => {
```
jessehon: Might be better to separate this out as it's own component

app/javascript/App/bundles/common/components/CoinTags.tsx
```diff
@@ -1,21 +1,36 @@
...
+      const isClickable = !!selectCoin
+      const onClickHandler = isClickable
```
jessehon: Would it be easier to just accept `onClick` as a prop? This way `CoinTags` do not need to know about selecting coins

app/javascript/App/bundles/common/components/CoinTags.tsx
jessehon: Sure, lets leave it for next time

PR #460, opened 2018-10-04
Improve query performance for NewsItem#serialize_news_items

anthotsang:
https://app.asana.com/0/683768070860536/836042970925470

### DIFF
--- a/app/models/news_coin_mention.rb
+++ b/app/models/news_coin_mention.rb
@@ -12,7 +12,7 @@ class NewsCoinMention < ApplicationRecord
     when 'machine'
       self.machine_tagged
     when 'combo'
-      self.human_tagged.or(self.machine_tagged)
+      self
     else
       raise "Invalid NEWS_COIN_MENTION_TAG_SCOPE environment variable"
     end
--- a/app/queries/news_items/with_filters.rb
+++ b/app/queries/news_items/with_filters.rb
@@ -14,7 +14,7 @@ module NewsItems
       # Apply FeedSources filter
       if feed_sources.blank?
         # Default feed sources
-        feed_sources = FeedSource.all
+        feed_sources = FeedSource.active
           .where.not(id: FeedSource.active.reddit)
           .where.not(id: FeedSource.active.twitter)
       end
--- a/app/queries/news_items/with_filters.rb
+++ b/app/queries/news_items/with_filters.rb
@@ -11,6 +11,15 @@ module NewsItems
     )
       result = relation
 
+      # Apply FeedSources filter
+      if feed_sources.blank?
+        # Default feed sources
+        feed_sources = FeedSource.all
+          .where.not(id: FeedSource.active.reddit)
+          .where.not(id: FeedSource.active.twitter)
+      end
+      result = result.where(feed_source: feed_sources)
+
       # Apply Coins filter
       if coins.present?
         filter_by_coins = true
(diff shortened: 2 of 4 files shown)


app/models/news_coin_mention.rb
```diff
@@ -12,7 +12,7 @@ def self.default_tagged
     when 'machine'
       self.machine_tagged
     when 'combo'
-      self.human_tagged.or(self.machine_tagged)
+      self
```
hanchang: Doh, this is so simple and clean ✨

db/migrate/20181004013839_add_composite_index_to_news_coin_mention.rb
```diff
@@ -0,0 +1,22 @@
...
+  disable_ddl_transaction!
+
+  def change
+    reversible do |dir|
+      dir.up do
+        # remove duplicates
+        execute <<-SQL
+          DELETE FROM news_coin_mentions AS T1 
+            USING news_coin_mentions AS T2 
+              WHERE T1.ctid < T2.ctid
```
hanchang: This is cool, didn't know about `ctid` in Postgres before!

db/migrate/20181004013839_add_composite_index_to_news_coin_mention.rb
```diff
@@ -0,0 +1,22 @@
+class AddCompositeIndexToNewsCoinMention < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def change
+    reversible do |dir|
+      dir.up do
+        # remove duplicates
+        execute <<-SQL
+          DELETE FROM news_coin_mentions AS T1
```
hanchang: @anthotsang Just a heads up that I personally prefer to separate schema migrations from data migrations; schema migrations live in `db/migrate` and data migrations live in `lib/tasks/data_migrations`.

The rationale for this is that it might be useful to have the data migration script separate from the schema migration so we can rerun it later if needed.

In this case though, I think it's fine to keep the data migration with this schema migration since it should be a one time task (as your unique index should prevent more duplicates in the future).

app/queries/news_items/with_filters.rb
```diff
@@ -11,6 +11,15 @@ def self.call(
     )
       result = relation
 
+      # Apply FeedSources filter
+      if feed_sources.blank?
+        # Default feed sources
+        feed_sources = FeedSource.all
```
hanchang: Yikes, just realized this should be `FeedSource.active` instead of `FeedSource.all`...

hanchang: @anthotsang @jessehon I'm gonna make this change, then please lemme know if anything else looks amiss, I think this PR is ready to go after the fix.

db/migrate/20181004013839_add_composite_index_to_news_coin_mention.rb
```diff
@@ -0,0 +1,22 @@
+class AddCompositeIndexToNewsCoinMention < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def change
+    reversible do |dir|
+      dir.up do
+        # remove duplicates
+        execute <<-SQL
```
hanchang: Actually we need to wrap this in a safety_assured { ... } block due to the Strong Migrations gem:
https://github.com/ankane/strong_migrations

PR #461, opened 2018-10-04
Switch to COINFI_NEW_PRICES_URL

hanchang:
https://app.asana.com/0/683768070860536/850578387606639

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -25,9 +25,6 @@
       <%= render partial: "components/flash" %>
       <%= yield %>
     </main>
-    <%= javascript_tag do %>
-      window.pricesURL = '<%= ENV.fetch('COINFI_PRICES_URL') %>';
-    <% end %>
     <script src="/tradingview/charting_library.min.js"></script>
     <%= javascript_pack_tag 'client-bundle' %>
     <%= render 'layouts/tracking_scripts' %>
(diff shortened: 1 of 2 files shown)


app/views/layouts/application.html.erb
```diff
@@ -25,9 +25,6 @@
       <%= render partial: "components/flash" %>
       <%= yield %>
     </main>
-    <%= javascript_tag do %>
-      window.pricesURL = '<%= ENV.fetch('COINFI_PRICES_URL') %>';
```
anthotsang: Is it okay to remove this? Maybe create a separate ticket for the cleanup?

Looks like it's used in API.js to set a remote base url via window.pricesURL
If we don't need it, it looks like localAPI.js and API.js could also be consolidated.
Seems like API.js is only used for CoinListContainer.tsx which at a glance appears to be related to the same data that the new url addresses.

hanchang: Thanks for the concern! I'm quite confident this can be removed; the only consumer of `API.js` is `CoinListContainer.tsx` like you mentioned, and all the API method calls in the CoinListContainer have the `remote` parameter set to false so the `if (remote) endpoint = window.pricesURL` line never gets invoked.

I will spin up another ticket to cleanup API.js though since it doesn't seem necessary but want to tackle that after we merge in your PR https://github.com/coinfi/coinfi/pull/445

PR #462, opened 2018-10-04
Improve webpack speed #840248853530078

jessehon:
https://app.asana.com/0/683768070860536/840248853530078


config/webpack/config/client.config.js
```diff
@@ -1,4 +1,6 @@
 const baseWebpackConfig = require('./base.config')
+const path = require('path')
```
anthotsang: is path used?

jessehon: Good catch, removed it

PR #463, opened 2018-10-04
[WIP] Setup alerting to detect stale values on daily, hourly, minutely price + volume ingest

anthotsang:
https://app.asana.com/0/683768070860536/843681325340141

Should be done but dblink step suddenly stopped providing results. Haven't been able to verify my volume logic.


lib/tasks/etl_db_check.rake
```diff
@@ -0,0 +1,138 @@
...
+            'url' => "https://hc-ping.com/67fffb49-c4d7-4e3f-a7cf-4688495c00d1"
+        },
+        {
+            'title' => "Top 10",
+            'query' => lambda { Coin.top(10).pluck(:coin_key, :ranking) },
+            'url' => "https://hc-ping.com/67fffb49-c4d7-4e3f-a7cf-4688495c00d1"
+        },
+        {
+            'title' => "Random 3",
+            'query' => lambda { Coin.where("ico_status=? OR ranking IS NOT NULL", 'listed')
```
hanchang: @anthotsang Actually this can probably just be `Coin.listed` since there's some metaprogramming in the `coin.rb` model that enabled this. The `ranking` is associated with CMC instead of CC which is where our data comes from so the `listed` scope is probably a better reflection of the data, apologies for the waffling!

lib/tasks/etl_db_check.rake
```diff
@@ -0,0 +1,138 @@
...
+                .order("RANDOM()").limit(3).pluck(:coin_key, :ranking) },
```
hanchang: What's the query performance for this `RANDOM()` call?

lib/tasks/etl_db_check.rake
```diff
@@ -0,0 +1,138 @@
+require 'net/http'
+require 'uri'
+
+namespace :etldb do
+    db_name = ENV['ETL_DB_NAME']
```
hanchang: Minor tweak requested - we use two space tabs in Ruby.

lib/tasks/etl_db_check.rake
```diff
@@ -0,0 +1,138 @@
...
+        {
+            'title' => "Random 3",
+            'query' => lambda { Coin.where("ico_status=? OR ranking IS NOT NULL", 'listed')
+                .order("RANDOM()").limit(3).pluck(:coin_key, :ranking) },
+            'url' => "https://hc-ping.com/67fffb49-c4d7-4e3f-a7cf-4688495c00d1"
+        }            
+    ]
+
+    desc "check for recent entries in specified table of etl database"
+    task :check, [:index] => :environment do |task, args|
```
jessehon: Can we add more supporting documentation or split this method up a bit. It's a bit hard to follow especially with 3 nested loops

lib/tasks/etl_db_check.rake
anthotsang: from the looks of it, the same as any ordering on a non-indexed column

lib/tasks/etl_db_check.rake
anthotsang: Added commenting. Let me know if it's still unclear.

PR #464, opened 2018-10-04
Use separate HotJar for /signals page #846448289962480

hanchang:
https://app.asana.com/0/683768070860536/846448289962480

### DIFF
--- a/app/views/layouts/_tracking_scripts.html.erb
+++ b/app/views/layouts/_tracking_scripts.html.erb
@@ -1,198 +1,191 @@
 <% if Rails.env.production? %>
-<!-- Heap Analytics -->
-<script type="text/javascript">
-  window.heap=window.heap||[],heap.load=function(e,t){window.heap.appid=e,window.heap.config=t=t||{};var r=t.forceSSL||"https:"===document.location.protocol,a=document.createElement("script");a.type="text/javascript",a.async=!0,a.src=(r?"https:":"http:")+"//cdn.heapanalytics.com/js/heap-"+e+".js";var n=document.getElementsByTagName("script")[0];n.parentNode.insertBefore(a,n);for(var o=function(e){return function(){heap.push([e].concat(Array.prototype.slice.call(arguments,0)))}},p=["addEventProperties","addUserProperties","clearEventProperties","identify","removeEventProperty","setEventProperties","track","unsetEventProperty"],c=0;c<p.length;c++)heap[p[c]]=o(p[c])};
-  heap.load('<%= ENV.fetch('HEAP_APP_ID') %>');
-  <% if current_user %>
-    heap.identify('<%= current_user.email %>');
-  <% end %>
-  window.addEventListener('load', function() {
-    if (typeof web3 !== 'undefined' && web3.currentProvider.isMetaMask) {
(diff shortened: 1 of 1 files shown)


app/views/layouts/_tracking_scripts.html.erb
```diff
@@ -78,17 +78,31 @@ twq('track','PageView');
 </script>
 <!-- End Twitter universal website tag code -->
 
+<% if request.original_fullpath.start_with? '/signals' %>
+<!-- Hotjar Tracking Code for https://www.coinfi.com/signals -->
```
jessehon: Personal preference but I actually feel it would be clearer to indent the contents of the `<% if ... %>`. It messes up the indentation in the generated html but I feel thats a better tradeoff if it means our code is more maintainable.

app/views/layouts/_tracking_scripts.html.erb
```diff
@@ -78,17 +78,31 @@ twq('track','PageView');
...
+<script>
+(function(h,o,t,j,a,r){
+  h.hj=h.hj||function(){(h.hj.q=h.hj.q||[]).push(arguments)};
+  h._hjSettings={hjid:1034618,hjsv:6};
```
jessehon: Is the hotjar id the only difference in these two blocks? Is it better to just customize that with a variable?

app/views/layouts/_tracking_scripts.html.erb
hanchang: Cool, I agree, will make the change, thanks!

app/views/layouts/_tracking_scripts.html.erb
hanchang: It is, great suggestion!

app/views/layouts/_tracking_scripts.html.erb
```diff
@@ -1,184 +1,191 @@
...
-<!-- End of Async Drift Code -->
-<% else %>
-<!-- UserSnap -->
-<script type="text/javascript">
-(function() { var s = document.createElement("script"); s.type = "text/javascript"; s.async = true; s.src = '//api.usersnap.com/load/566ea3b0-5719-4a7c-999f-b7f6604930c2.js';
-var x = document.getElementsByTagName('script')[0]; x.parentNode.insertBefore(s, x); })();
-</script>
-<% end %>
+  <!--
+  Only show Mouseflow on landing pages, otherwise the recording quota
```
jessehon: This might be better as a rails comment. Since these html comments will get rendered to page source

app/views/layouts/_tracking_scripts.html.erb
```diff
@@ -1,184 +1,191 @@
...
-    appId: "91b756ae-dfc8-45ab-b96e-93aa1bf9d474",
-    allowLocalhostAsSecureOrigin: true,
-    autoRegister: true,
-    notifyButton: {
-      enable: false /* Set to false to hide */
-    }
-  }]);
-</script>
+  <!--
+  Only show Drift on Signals page, otherwise show Usersnap bug report;
```
jessehon: Similar to above, might be better as a rails comment

PR #466, opened 2018-10-05
Write test to ensure TradingView chart loads

anthotsang:
https://app.asana.com/0/683768070860536/841232351150904


hanchang:
Let's chat about this as a team on Monday, Oct 15 when you're back @anthotsang to settle on which paradigm we want to use for testing moving forward.

In the meantime let's leave this PR open for now.

Gemfile.lock
```diff
@@ -17,15 +17,15 @@ GIT
 
 GIT
   remote: https://github.com/nickcharlton/administrate-field-nested_has_many.git
-  revision: 10eefee002395a55fccf4a59c318d7da38df22e6
+  revision: 47a336e08bec01a51048ce0f6b99c23c7815dbd7
```
hanchang: Hm, I can't tell if the changes in this file are due to the specification of the `selenium-webdriver` gem to lock to version `>= 3.14.0` or if this is a wholesale upgrade of all of our Ruby gems, can you confirm please @anthotsang ?

PR #467, opened 2018-10-05
Setup system test environment #852574080062597

jessehon:
https://app.asana.com/0/683768070860536/852574080062597


test/factories/coins.rb
```diff
@@ -0,0 +1,54 @@
+require 'faker'
+
+FactoryBot.define do
+  factory :coin do
+    sequence(:name) { |n| "#{Faker::CryptoCoin.coin_name}#{n}" }
```
hanchang: I can't believe `Faker::CryptoCoin` is a thing haha

jessehon: Haha, it's not really that useful though

PR #469, opened 2018-10-09
Signal reservation updates #852614871284588

jessehon:
https://app.asana.com/0/683768070860536/852614871284588


app/javascript/App/bundles/signalsPage/SignalReservationForm.tsx
```diff
@@ -0,0 +1,355 @@
...
+    axios
+      .patch(SIGNALS_RESERVATION_URL, this.state.formData, { headers })
+      .then(() => {
+        this.setState((state) => ({
+          activeStep: state.activeStep + 1,
+        }))
+      })
+      .catch((error) => {
+        // TODO: Handle error by showing some sort of message.
+        alert(error)
```
jessehon: Are we happy to leave this as an `alert`?

app/javascript/App/bundles/signalsPage/SignalReservationForm.tsx
```diff
@@ -0,0 +1,355 @@
...
+    return haveValidationsPassed
+  }
+
+  public handleNext = () => {
+    if (!this.validateStep(this.state.activeStep)) {
+      return
+    }
+
+    const headers = {
+      'X-CSRF-Token': document
```
jessehon: It is better to pass this in via a prop than try to assume there would always be this element. There is an example of how I do this in the `NavUser` component and the logout button

app/javascript/App/bundles/signalsPage/SignalReservationForm.tsx
hanchang: Yeah, I was going to use Material UI popup but figured it's overkill for now, but we should go back and tidy it up later.

jessehon: Sure

PR #473, opened 2018-10-11
Signals Telegram Bot API #856546340183783

jessehon:
https://app.asana.com/0/683768070860536/856546340183783

Added two new environment variables

- SIGNALS_TELEGRAM_BOT_API_USERNAME
- SIGNALS_TELEGRAM_BOT_API_PASSWORD


app/forms/signals_telegram_bot_registration_form.rb
```diff
@@ -0,0 +1,37 @@
...
+  def persist
+    user.token_sale[:signals_telegram_bot_chat_id] = chat_id
+    user.token_sale[:signals_telegram_bot_started_at] = started_at
+    user.save
+  end
+
+  protected
+
+  def user
+    @user ||= User.where("(token_sale->>'telegram_username') = ?", telegram_username).first
```
hanchang: Are we okay to just throw an error if no user is found with the provided `telegram_username`?

jessehon: Does it throw an error? It shouldn't and would just return `nil` if not found

hanchang: Sorry, this line won't throw an error but I'm thinking the `#persist` method would subsequently throw since we'd be calling `nil.token_sale` in the case of no user found.

config/routes.rb
```diff
@@ -51,6 +51,11 @@
       get 'watchlist', on: :collection
     end
 
+    resources :signals_telegram_bot, only: [] do
```
hanchang: Maybe `namespace` instead of `resources ... only: []` would be more readable?

app/forms/signals_telegram_bot_registration_form.rb
jessehon: Ah, the logic behind this is in the parent `Patterns::Form` class. The `persist` method won't run until the validations pass and there is a validation for `validates : user, presence: true...`

hanchang: Ah very cool, thanks for the explanation!

PR #475, opened 2018-10-15
do not overwrite available_supply for Coins, let ETL database populate this field

anthotsang:
https://app.asana.com/0/683768070860536/812976565167603

### DIFF
--- a/app/models/coin.rb
+++ b/app/models/coin.rb
@@ -154,7 +154,7 @@ class Coin < ApplicationRecord
   def market_info market_data = nil
     data = market_data || live_market_data.dup
     data["24h_volume_usd"] = humanize(data["24h_volume_usd"], '$') if data["24h_volume_usd"]
-    if self.available_supply then
+    if self.available_supply
       data["available_supply"] = humanize(self.available_supply)
     elsif data["available_supply"]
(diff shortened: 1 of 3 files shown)


app/models/coin.rb
```diff
@@ -154,7 +154,11 @@ def live_market_data
   def market_info market_data = nil
     data = market_data || live_market_data.dup
     data["24h_volume_usd"] = humanize(data["24h_volume_usd"], '$') if data["24h_volume_usd"]
-    data["available_supply"] = humanize(data["available_supply"]) if data["available_supply"]
+    if self.available_supply then
```
hanchang: Whoa, didn't realize Ruby had a `then` keyword... very cool!

I don't quite see the value of its usage here though, shall we remove?

I think the only usage I think makes sense is a single line `if/else` statement, i.e.

```
if (1==2) then puts "Math doesn't work" else puts "Math works!" end
```

Credit to:
https://stackoverflow.com/questions/5659360/why-should-you-avoid-the-then-keyword-in-ruby

anthotsang: Yeah, seems to be a way to fix ambiguous if statements that's not needed here. Thanks.

PR #476, opened 2018-10-15
Prevent double sending Signals RSVP email #864344982710644

hanchang:
https://app.asana.com/0/683768070860536/864344982710644


anthotsang:
Looks like this page would still be susceptible to double clicking the button?

hanchang:
> Looks like this page would still be susceptible to double clicking the button?

Yeah we decided that's okay for now

PR #480, opened 2018-10-16
Improving the coin search bar filter logic

anthotsang:
https://app.asana.com/0/683768070860536/854461086638935

### DIFF
--- a/app/controllers/api/coins_controller.rb
+++ b/app/controllers/api/coins_controller.rb
@@ -23,9 +23,17 @@ class Api::CoinsController < ApiController
     if params[:coinSlugs].present? 
       coins = Coin.where(slug: params[:coinSlugs])
     elsif params[:name].present?
-      coins = Coin.where('upper(concat(symbol,\' \', name)) like upper(?)', "%#{params[:name]}%")
-                  .order('length(name) asc')
-                  .limit(10)
+      coins = Coin.find_by_sql("
+        SELECT *, CASE
+            WHEN UPPER(symbol) = UPPER('#{params[:name]}') THEN 1
+            WHEN UPPER(name) = UPPER('#{params[:name]}') THEN 1
+            ELSE 3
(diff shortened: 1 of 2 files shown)


app/models/coin.rb
```diff
@@ -110,12 +110,12 @@ def summary
...
 
-    most_common_feed_source_name = self.most_common_feed_source.name
-    most_common_news_category_name = self.most_common_news_category.name
+    most_common_feed_source_name = self.most_common_feed_source&.name
+    most_common_news_category_name = self.most_common_news_category&.name
     result << %W[
       The most common news source covering #{self.name} is #{most_common_feed_source_name} and the
       most common news category is #{most_common_news_category_name}.
-    ]
+    ] if most_common_feed_source_name != nil and most_common_news_category_name != nil
```
hanchang: Thanks for the catch @anthotsang !

Jesse has an existing fix that's already deployed though, so let's roll this one back please to handle the merge conflict.

app/controllers/api/coins_controller.rb
```diff
@@ -23,9 +23,17 @@ def search_by_params
     if params[:coinSlugs].present? 
       coins = Coin.where(slug: params[:coinSlugs])
     elsif params[:name].present?
-      coins = Coin.where('upper(concat(symbol,\' \', name)) like upper(?)', "%#{params[:name]}%")
-                  .order('length(name) asc')
-                  .limit(10)
+      coins = Coin.find_by_sql("
```
hanchang: @anthotsang Just curious about how the query performance on this?

anthotsang: Seems to be similar to the previous query. The EXPLAIN() for both has the same plan/costs.

app/models/coin.rb
anthotsang: Reverted

PR #493, opened 2018-10-22
Create homepage top section banner

anthotsang:
https://app.asana.com/0/683768070860536/871520362652036

### DIFF
--- a/app/javascript/App/bundles/common/components/Banner.tsx
+++ b/app/javascript/App/bundles/common/components/Banner.tsx
@@ -1,0 +1,81 @@
+import * as React from 'react'
+import { withStyles, createStyles } from '@material-ui/core/styles'
+import { Button, Paper, Typography, Grid } from '@material-ui/core'
+
+const styles = (theme) =>
+  createStyles({
+    mainFeaturedPost: {
+      backgroundColor: '#253640',
+      color: '#d7d7d7',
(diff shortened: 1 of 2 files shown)


app/javascript/App/bundles/common/components/Banner.tsx
```diff
@@ -0,0 +1,81 @@
...
+          className={classes.subheader}
+          color="inherit"
+          paragraph={true}
+        >
+          Uncover buy and sell opportunities through data science backed
+          blockchain analytics.
+        </Typography>
+      </Grid>
+      <Grid item={true} md={true}>
+        <Button
```
hanchang: @jlei523 We probably want to instrument Heap yeah? Can you help provide @anthotsang with the appropriate instrumentation docs and data attribute for this?

jlei523: @anthotsang Here's the documentation: https://app.nuclino.com/CoinFi/Product/Instrumenting-Heap-Analytics-Events-via-data-attributes-e800f8d2-414b-47cd-bccd-20d29c1438ce

For this event, let's name it "homepage-click-signals-cta"

anthotsang: @jlei523 Thanks. I've added the attribute. Based on the above documentation, I assume I don't need to add the event on Heap, but please let me know if I should do that too.

PR #501, opened 2018-10-23
Use env var to set Signals staking amount #875373581992587

hanchang:
https://app.asana.com/0/683768070860536/875373581992587


app/views/signals/reservation.html.erb
```diff
@@ -22,7 +25,7 @@
       <h2 class="reservation-section__title">FAQ</h2>
 
       <div class="signal-faqs">
-        <%= react_component_with_jss "SignalFaqPanel",
+        <%= react_component "SignalFaqPanel",
```
anthotsang: I'm assuming changes to react_component_with_jss and prerender: true are unintentional?

hanchang: Thanks for checking @anthotsang, this was intentional; I'm temporarily rolling back since the SSR JSS is causing some visual weirdness and we need to get this out ASAP.

PR #503, opened 2018-10-23
Create news homepage widget

anthotsang:
https://app.asana.com/0/683768070860536/871520362652044


app/controllers/api/news_controller.rb
```diff
@@ -88,4 +59,48 @@ def serialized(obj)
       formatted_data = format_item.call(data)
     end
   end
+
+  def process_news_feed(params)
```
jessehon: Hmmm, the name of this method makes me feel it would produce side effects (e.g. save actions). Maybe `apply_news_feed_filters(news_items, params)`

app/controllers/api/news_controller.rb
```diff
@@ -1,60 +1,31 @@
...
-        coins = Coin.where(slug: params[:coinSlugs])
+      news_items = nil
+
+      if params[:frontPage].present? # For Front page (public; no feature flag)
+        news_items = NewsItem.published.front_page
+          .includes(:coins, :news_categories)
+          .order_by_published
+          .limit(5)
+      else # For newsfeed (keep feature flag)
+        has_news = detect_news_feature
```
jessehon: I think `has_news` sounds like it would be a `boolean` but in this case is actually a json_response. Maybe we can change the name of this variable to be a little clearer

app/models/news_item.rb
```diff
@@ -19,6 +19,7 @@ class NewsItem < ApplicationRecord
   scope :tagged, -> { where(is_human_tagged: true) }
   scope :twitter, -> { where(feed_source: FeedSource.twitter) }
   scope :order_by_published, -> (order = nil) { order(feed_item_published_at: order || :desc) }
+  scope :front_page, -> { where(feed_source: FeedSource.coindesk.or(FeedSource.cointelegraph)) }
```
jessehon: Maybe worth moving this into a separate `Query` in `app/queries` since this specific to a particular view

anthotsang: You're right, it probably won't be recycled. I would propose moving it to the query on the controller instead then? Setting up a new model in app/queries for it seems like it might be overkill?

jessehon: Sure that works too

PR #505, opened 2018-10-24
Integrate Skylight #879945295176565

hanchang:
https://app.asana.com/0/683768070860536/879945295176565


hanchang:
Note: According to the docs, the gem should be installed across environments and not just in the `production` group, although I'm not sure why.

PR #510, opened 2018-10-24
Update staked token amount property name

jessehon:


app/forms/signals_telegram_bot_registration_form.rb
```diff
@@ -24,13 +24,13 @@ def user
...
-    if staked_token_amount < self.class::REQUIRED_MIN_STAKED_TOKEN_AMOUNT
+  def validate_user_staked_cofi_amount
+    staked_cofi_amount = user.token_sale.fetch('staked_cofi_amount', 0)
+    if staked_cofi_amount < self.class::REQUIRED_MIN_STAKED_COFI_AMOUNT
       errors.add(
         :user,
-        :min_staked_token_amount,
-        message: "has not staked more than #{self.class::REQUIRED_MIN_STAKED_TOKEN_AMOUNT} tokens"
+        :min_staked_cofi_amount,
+        message: "has not staked more than #{self.class::REQUIRED_MIN_STAKED_COFI_AMOUNT} cofi tokens"
```
hanchang: If this is user facing we should capitalize `COFI`

jessehon: not user facing actually, but going to capitalize it anyways

PR #511, opened 2018-10-25
Improve telegram bot query performance #881853263276622

jessehon:
https://app.asana.com/0/683768070860536/881853263276622


hanchang:
We can go ahead and merge into `develop` and then `master` since there's nothing on develop right now.

PR #512, opened 2018-10-25
Render Historical Data table on Coin Detail page

anthotsang:
https://app.asana.com/0/683768070860536/866838733053116

### DIFF
--- a/app/javascript/App/bundles/CoinShow/HistoricalData.tsx
+++ b/app/javascript/App/bundles/CoinShow/HistoricalData.tsx
@@ -1,0 +1,211 @@
+import * as React from 'react'
+import * as _ from 'lodash'
+import {
+  Table,
+  TableHead,
+  TableBody,
+  TableRow,
+  TableCell,
+  Typography,
+  Toolbar,
+} from '@material-ui/core'
+import { withStyles, createStyles } from '@material-ui/core/styles'
+import LoadingIndicator from '../common/components/LoadingIndicator'
+import * as moment from 'moment'
+import { Moment } from 'moment'
+import DateRangeSelect from './DateRangeSelect'
+
+enum STATUSES {
+  INITIALIZING = 'INITIALIZING',
+  LOADING = 'LOADING',
+  READY = 'READY',
+}
+
+interface RawPriceData {
+  coin_key: string
+  to_currency: string
+  time: string
+  open: number
+  high: number
+  low: number
+  close: number
+  volume_from: number // volume in present currency
+  volume_to: number // volume in usd (or to_currency?)
+}
+
+interface PriceData {
+  time: string
+  open: string
+  high: string
+  low: string
+  close: string
+  volume: string
+  formattedTime: string
+  datetime: Moment
+  marketCap: string
+}
+
+interface Props {
+  classes: any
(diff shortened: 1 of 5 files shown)


app/javascript/App/bundles/CoinShow/HistoricalData.tsx
```diff
@@ -0,0 +1,211 @@
...
+  createStyles({
+    table: {},
+    toolbar: {},
+    grow: {
+      flexGrow: 1,
+    },
+    picker: {},
+  })
+
+class HistoricalData extends React.Component<Props, State> {
```
hanchang: Minor nitpick - could we rename this to `HistoricalPriceDataTable` for clarity?

app/javascript/App/bundles/CoinShow/HistoricalPriceDataTable.tsx
```diff
@@ -0,0 +1,211 @@
...
+      data: initialSortedData || [],
+      currency,
+    }
+  }
+
+  public sortDataFunc(x: PriceData, y: PriceData) {
+    return y.datetime.diff(x.datetime)
+  }
+
+  public formatPrice(price: number, decimal: number = 2): string {
```
hanchang: Since some cryptocurrencies are valued at sub-cent prices per token, can we go out to 6 decimal places?

While we're at it let's prepend the dollar sign as well, thanks!

app/javascript/App/bundles/CoinShow/HistoricalPriceDataTable.tsx
```diff
@@ -0,0 +1,211 @@
...
+    return price.toLocaleString('en-US', {
+      maximumFractionDigits: decimal,
+    })
+  }
+
+  public parseData(d: RawPriceData): PriceData {
+    const datetime = moment.utc(d.time)
+    const formattedTime = datetime.format('MMM DD, YYYY')
+    const marketCap = this.formatPrice(
+      Math.round(this.props.availableSupply * d.close),
```
hanchang: Hm, sorry for missing this requirement - if Available Supply is missing we should just hide the Market Cap column since it'll be empty.

app/javascript/App/bundles/CoinShow/HistoricalPriceDataTable.tsx
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  public onDateChangeHandler = (startDate: string, endDate: string) => {
+    this.setState({
+      start: startDate ? moment(startDate) : null,
+      end: endDate ? moment(endDate) : null,
+    })
+  }
+
+  public componentDidMount() {
```
hanchang: @anthotsang Is this method going to be used in the future?

anthotsang: I was setting the component up assuming that it might not have the data up front. But I haven't actually bothered filling out all the async parts yet since it's not needed at this time. I can remove it for now if you'd like.

hanchang: I'm okay to leave it for now as long as we remember, thanks for the heads up!

PR #525, opened 2018-10-29
Add index to `NewsItems.title` database column

anthotsang:
https://app.asana.com/0/683768070860536/881853263276627

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -306,7 +306,6 @@ ActiveRecord::Schema.define(version: 20181029150802) do
     t.index ["coin_id"], name: "index_feed_sources_on_coin_id"
     t.index ["feed_type"], name: "index_feed_sources_on_feed_type"
     t.index ["feed_url"], name: "index_feed_sources_on_feed_url", unique: true
-    t.index ["id", "is_active", "feed_type"], name: "compound_index"
     t.index ["is_active"], name: "index_feed_sources_on_is_active"
     t.index ["name"], name: "index_feed_sources_on_name", unique: true
   end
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -306,6 +306,7 @@
     t.index ["coin_id"], name: "index_feed_sources_on_coin_id"
     t.index ["feed_type"], name: "index_feed_sources_on_feed_type"
     t.index ["feed_url"], name: "index_feed_sources_on_feed_url", unique: true
+    t.index ["id", "is_active", "feed_type"], name: "compound_index"
```
hanchang: Hm this looks new, what's the background?

db/schema.rb
```diff
@@ -342,6 +343,13 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "market_caps", force: :cascade do |t|
```
hanchang: Let's stash these lines for this particular PR please.

db/schema.rb
anthotsang: Looks like it's from a while ago but the schema.db was never updated.

https://github.com/coinfi/coinfi/blob/develop/db/migrate/20180928070531_add_is_active_index_to_feed_sources.rb

db/schema.rb
anthotsang: Oops sorry, will clean that up.

db/schema.rb
hanchang: Huh weird, the line below (310) should be the output for that migration though...

Let's go ahead and remove still unless I'm missing something?

anthotsang: Ah, you're right. I think this comes from a compound index I manually added when trying to optimize news items. :cry:

PR #527, opened 2018-10-30
Automate transaction confirmation #879942592574994

jessehon:
https://app.asana.com/0/683768070860536/879942592574994


app/forms/signals_telegram_bot_registration_form.rb
```diff
@@ -8,7 +7,7 @@ class SignalsTelegramBotRegistrationForm < Patterns::Form
 
   validates :telegram_username, presence: true
   validates :user, presence: true, :if => proc { |f| f.telegram_username.present? }
-  validate :validate_user_staked_cofi_amount, :if => proc { |f| f.user.present? }
+  validate :validate_user_access, :if => proc { |f| f.user.present? }
```
hanchang: Can we reorder this to the end of the `validates` lines to make it more clear that this particular validation is referring to a method instead of a model field?

jessehon: Sure, moving it definitely a lot clearer

app/forms/signals_telegram_bot_registration_form.rb
```diff
@@ -24,13 +23,36 @@ def user
     @user ||= User.where("(token_sale->>'telegram_username') ILIKE ?", telegram_username).first
   end
 
+  def validate_user_access
+    signals_access_override = user.token_sale.fetch('signals_access_override', nil)
+
+    # Handle `signals_access_override` values
+    unless signals_access_override.nil?
+      if signals_access_override == true
```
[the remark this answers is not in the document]
jessehon: I actually wanted to avoid just truthy values and instead throw an error if it is not exactly boolean. I think it just helps avoid any `"false"` strings from accidentally returning `true`

PR #528, opened 2018-10-30
Install rack-timeout gem #887621176139248

hanchang:
https://app.asana.com/0/683768070860536/887621176139248


hanchang:
I'm gonna wait to merge this until we let the Rack Attack installation settle for a bit, to reduce the number of variables for easier debugging in case issues crop up.

PR #531, opened 2018-10-30
Bugfix: News filtering by coin

hanchang:


app/queries/news_items/with_filters.rb
```diff
@@ -31,7 +31,7 @@ def self.call(
         result = result.where(feed_source: feed_sources)
       end
 
-      news_coin_mentions = NewsCoinMention.default_tagged
+      news_coin_mentions = NewsCoinMention.default_tagged.where(coin: coins)
```
anthotsang: Ah, sorry about that. I think there where should be on `.where("news_coin_mentions.id IN (?)", news_coin_mentions.where(coin: coins).select(:id))`

hanchang: Gotcha, good catch!

PR #541, opened 2018-11-02
Integrate newly styled price charts

anthotsang:
https://app.asana.com/0/683768070860536/881853263276643

### DIFF
--- a/app/javascript/App/bundles/common/components/CoinCharts/PriceGraph/chartOptions.js
+++ b/app/javascript/App/bundles/common/components/CoinCharts/PriceGraph/chartOptions.js
@@ -1,3 +1,7 @@
+import _ from 'lodash'
+import moment from 'moment'
+import timeago from 'timeago.js'
+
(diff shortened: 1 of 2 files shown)


app/javascript/App/bundles/common/components/CoinCharts/PriceGraph/chartOptions.js
```diff
@@ -1,3 +1,7 @@
+import _ from 'lodash'
+import moment from 'moment'
+import timeago from 'timeago.js'
```
jessehon: Do we need `timeago` if we already have `moment`? moment has a `fromNow()` that I think does the same thing https://momentjs.com/docs/#/displaying/fromnow/

anthotsang: timeago was what was already being used in news
I'm fine with moving away from it though.

PR #542, opened 2018-11-02
Manage watchlist from telegram bot #887812181500958

jessehon:
https://app.asana.com/0/683768070860536/887812181500958


lib/tasks/data_migrations/09_migrate_signals_telegram_data_to_signals_telegram_users.rake
```diff
@@ -0,0 +1,23 @@
+namespace :data_migrations do
+  desc 'Migrate signals telegram data to `signals_telegram_users`'
+  task :migrate_signals_telegram_data_to_signals_telegram_users => :environment do
+    ActiveRecord::Base.transaction do
+      users = User.where("(token_sale->>'signals_telegram_bot_chat_id') IS NOT NULL")
+      users.find_each do |user|
```
hanchang: Want to use `batch_process` for this? Optional in my book.

jessehon: `batch_process` doesn't run in a transaction, meaning it could leave some users migrated and some not. If there is a problem with migration, I think it may be easier to have all or nothing when resolving any issues.

hanchang: Ah yes definitely, thanks!

PR #549, opened 2018-11-06
Skip Devise Trackable for API calls

hanchang:
https://app.asana.com/0/683768070860536/892130265558824


app/controllers/api/coins_controller.rb
```diff
@@ -54,8 +54,10 @@ def toplist
   end
 
   def watchlist
-    coins = current_user.watchlist.coins.order(:ranking)
-    respond_success coinlist_serializer(coins)
+    if current_user
```
anthotsang: I guess this could be made to fail more gracefully for logged out users.

PR #550, opened 2018-11-06
Update Signals Landing Page Messaging

hanchang:
https://app.asana.com/0/683768070860536/882266191521956

### DIFF
--- a/app/javascript/App/bundles/signalsPage/SignalExamplePanel.tsx
+++ b/app/javascript/App/bundles/signalsPage/SignalExamplePanel.tsx
@@ -38,6 +38,11 @@ const styles = (theme) =>
     panelExpandIcon: {
       right: 0,
     },
+    activeText: {
+      color: '#2faeed',
+      fontWeight: 'bold',
+      paddingBottom: '0 !important',
+    },
     tableRow: {
       // Table striping is enabled for all `<tr>`s by default with a global and an extremely
       // specific css rule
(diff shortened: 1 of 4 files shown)


app/views/signals/index.html.erb
```diff
@@ -124,6 +124,7 @@
       <div class="col-sm-6 signals-section__col">
         <div class="signals-section-content signals-section-content--xs-last">
           <div class="signal-example-panels">
+            <div class="b sky-blue">Active! Our beta users are trading with this signal.</div>
```
anthotsang: I wonder if it would be better to put this into the component using a prop of active: true ?
I don't know if it will be applied to other signals later, but the placement is a bit confusing.

hanchang: Yeah... it'll almost definitely be used for other signals. I tried that approach and was unsuccessful, but just realized it's likely because I was only running the server rendering Webpack build - will give it another shot with the client side.

app/javascript/App/bundles/signalsPage/SignalExamplePanel.tsx
```diff
@@ -72,6 +77,7 @@ interface Props extends DeviceContextType {
   title: string
   buyDescription?: string
   sellDescription?: string
+  active?: boolean
```
hanchang: I personally prefer boolean variables to be prefixed with "is", i.e. `isActive` but I think it's fine here.

hanchang:
🥇 Looks great @anthotsang, thanks for the help!

PR #551, opened 2018-11-07
Cache default News#index and News#show query

anthotsang:
https://app.asana.com/0/683768070860536/897417176249795


app/controllers/webhooks/websubs_controller.rb
```diff
@@ -13,6 +15,19 @@ def superfeedr_ingest
 
     puts "Received #{items.count} NewsItems from SuperFeedr."
 
+    if items.count > 0
+      Rails.cache.write("news") do
+        distribute_reads(max_lag: MAX_ACCEPTABLE_REPLICATION_LAG, lag_failover: true) do
+          news_items = default_news_query
+          if default_news_query.empty?
```
hanchang: This is a nitpick but calling `default_news_query` twice would send two SQL queries, can we just do

```
news_items = default_news_query
if news_items.empty?
  news_items = backup_default_news_query
end
```

anthotsang: make sense. thanks!

PR #556, opened 2018-11-08
Integrate pgHero #900264321199832

hanchang:
https://app.asana.com/0/683768070860536/900264321199832


app.json
```diff
@@ -71,7 +73,7 @@
     "heroku-redis:hobby-dev",
     "papertrail",
     "rollbar",
-    "scout",
```
anthotsang: no longer using scout?

hanchang: Yeah, Librato has more metrics and Skylight handles the slow query analysis better.

PR #557, opened 2018-11-08
Allow user to set and persist default currency

anthotsang:
https://app.asana.com/0/683768070860536/892261161134131


app/javascript/App/bundles/CoinShow/FundamentalsList.tsx
```diff
@@ -54,58 +57,66 @@ class FundamentalsList extends React.Component<Props, {}> {
     const hasIcoEnded = ico_end_epoch < moment().unix()
 
     return (
-      <Grid container={true} direction="column">
-        {hasStartDate && (
-          <>
-            <Grid item={true} className={classes.title}>
-              Start date
+      <CurrencyContext.Consumer>
+        {({ currency, currencyRate }: CurrencyContextType) => {
```
jessehon: Is this `currency` and `currencyRate` used in this component?

app/javascript/App/bundles/CoinShow/HistoricalPriceDataTable.tsx
```diff
@@ -101,36 +99,36 @@ class HistoricalPriceDataTable extends React.Component<Props, State> {
...
     const initialStatus = statusIsReady ? STATUSES.READY : STATUSES.INITIALIZING
-    const currency = initialData
-      ? _.get(initialData, ['0', 'to_currency'])
-      : undefined
 
     this.state = {
       status: initialStatus,
       data: initialSortedData || [],
-      currency,
+      rawData: initialData || [],
```
jessehon: Is it better to call this `initialRawData` to make the distinction between this `rawData` field and `data`

anthotsang: Since it's state, I can see the rawData being updated asynchronously at some point and won't necessary only be initialRawData.

app/javascript/App/bundles/CoinShow/FundamentalsList.tsx
anthotsang: It was previously, but they took out all the fundamentals that use it. I was going to rip it out but there's still on-going discussion of adding some price-related values back in so I would prefer to leave it even if it's useless right now.

app/javascript/App/bundles/CoinShow/HistoricalPriceDataTable.tsx
jessehon: Put the comment on the wrong spot really, I meant renaming the `prop.initialData` to `prop.initialRawData`

app/javascript/App/bundles/CoinShow/HistoricalPriceDataTable.tsx
```diff
@@ -141,17 +139,25 @@ class HistoricalPriceDataTable extends React.Component<Props, State> {
   }
 
   public parseData(d: RawPriceData): PriceData {
```
jessehon: You can do `parseData = (d: RawPriceData): PriceData => {` to avoid explicitly binding to `this`

app/javascript/App/bundles/CoinShow/HistoricalPriceDataTableWrapper.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react'
+import HistoricalPriceDataTable from './HistoricalPriceDataTable'
+import CurrencyContext, {
+  CurrencyContextType,
+} from '~/bundles/common/contexts/CurrencyContext'
+
+const HistoricalPriceDataTableWrapper = (props) => (
+  <CurrencyContext.Consumer>
```
jessehon: This Consumer seems to be used quite a bit. I wonder if its worth creating a separate HOC for it to feed the props into the wrapped component. Similar to `withStyles()`. It would save having to create a new component like this.

app/javascript/App/bundles/CoinShow/InfoBar.tsx
```diff
@@ -138,112 +140,126 @@ class InfoBar extends React.Component<Props, {}> {
...
-          <Grid item={true} className={classes.watchButtonContainer}>
-            <Icon
-              name="star"
-              solid={true}
-              className={classnames(
-                classes.watchButton,
-                isWatched ? classes.watchedButton : classes.unwatchedButton,
-              )}
-              onClick={watchCoinHandler}
+      <CurrencyContext.Consumer>
```
jessehon: You should be able to use your new `withCurrency` HOC in all the areas that need `CurrencyContext.Consumer`

app/javascript/App/bundles/common/components/CoinCharts/PriceGraph/index.js
```diff
@@ -37,6 +57,85 @@ class PriceGraph extends Component {
...
+      this.state.chart.series[0].name = priceLabel
+      this.state.chart.yAxis[0].setTitle({
+        text: priceLabel,
+      })
+      this.state.chart.series[1].name = volumeLabel
+      this.state.chart.yAxis[1].setTitle({
+        text: volumeLabel,
+      })
+    } catch (e) {
+      console.error(e)
```
jessehon: Is there a reason we expect `setCurrency` to potentially throw an error?

app/javascript/App/bundles/common/components/CoinCharts/TradingViewChart/index.js
```diff
@@ -62,7 +75,50 @@ export default class TradingViewChart extends Component {
...
+          this.resetHandler()
+          chart.resetData()
+        }
+      } catch (e) {
+        console.error(e)
+      }
+    }
+
+    if (prevProps.currency !== this.props.currency) {
+      // currency change logic here
```
jessehon: Is this a placeholder for functionality to be implemented next time?

app/javascript/App/bundles/common/components/CoinCharts/TradingViewChart/index.js
```diff
@@ -62,7 +75,50 @@ export default class TradingViewChart extends Component {
...
+      const datafeed = new Datafeed(priceData, priceDataHourly)
+
+      try {
+        if (this.tvWidget && this.tvWidget.chart && this.resetHandler) {
+          const chart = this.tvWidget.chart()
+          this.resetHandler()
+          chart.resetData()
+        }
+      } catch (e) {
+        console.error(e)
```
jessehon: Similar to above, is there a reason for this `try`/`catch`

app/javascript/App/bundles/CoinShow/InfoBar.tsx
anthotsang: Yeah I'm aware, but didn't really feel like converting them all since it wasn't really necessary.

jessehon: I guess we can just slowly migrate those components across

PR #562, opened 2018-11-08
Limit watchlist signal token message to only ERC20 tokens #902471154637928

jessehon:
https://app.asana.com/0/683768070860536/902471154637928

### DIFF
--- a/app/controllers/api/signals_telegram_bot/signals_telegram_subscriptions_controller.rb
+++ b/app/controllers/api/signals_telegram_bot/signals_telegram_subscriptions_controller.rb
@@ -66,9 +66,14 @@ class Api::SignalsTelegramBot::SignalsTelegramSubscriptionsController < Api::Sig
   end
 
   def serialize_signals_telegram_subscription(signals_telegram_subscription)
-    coin_json = signals_telegram_subscription.coin.as_json(
-      only: %i[id name symbol slug token_type]
-    )
+    coin_json = signals_telegram_subscription.coin
+      .as_json(
+        only: %i[id name symbol slug]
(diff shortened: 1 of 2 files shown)


app/controllers/api/signals_telegram_bot/signals_telegram_subscriptions_controller.rb
```diff
@@ -66,9 +66,14 @@ def create_params
...
   def serialize_signals_telegram_subscription(signals_telegram_subscription)
-    coin_json = signals_telegram_subscription.coin.as_json(
-      only: %i[id name symbol slug]
-    )
+    coin_json = signals_telegram_subscription.coin
+      .as_json(
+        only: %i[id name symbol slug]
+      )
+      .merge({
+        is_erc20: signals_telegram_subscription.coin.is_erc20?
```
hanchang: I think in this particular case it makes more sense to just do `signals_telegram_subscription.coin.eth_address.present?` because if we don't have the ETH address of the token, even if it is an ERC-20 token we won't be able to provide signals for it.

Unfortunately that'll probably necessitate some renaming of variables for clarity.

jessehon: How about `is_supported_erc20`?

hanchang: I've gone ahead and spun up a ticket on the Data Integrity board though to see if any tokens should have an `eth_address` but currently do not...

https://app.asana.com/0/794593063903042/902331431951580

PR #563, opened 2018-11-08
Remove unused database indexes #902302351759198

hanchang:
https://app.asana.com/0/683768070860536/902302351759198


db/migrate/20181108032049_remove_unused_indexes.rb
```diff
@@ -0,0 +1,29 @@
+class RemoveUnusedIndexes < ActiveRecord::Migration[5.1]
+  def change
+    remove_index :coins, column: :volume24 # name: "index_coins_on_volume24"
```
jessehon: Why not pass the name in as a parameter?

hanchang: Great question - passing in the `column` allows the migration to be rolled back automatically, otherwise the migration is one way only.

PR #564, opened 2018-11-08
Set pg statement timeout in database.yml #900659722533378

hanchang:
https://app.asana.com/0/683768070860536/900659722533378


config/database.yml
```diff
@@ -34,6 +34,8 @@ default: &default
   adapter: postgresql
   url: <%= ENV["DATABASE_URL"] %>
   <% end %>
+  variables:
+    statement_timeout: <%= ENV.fetch("POSTGRES_STATEMENT_TIMEOUT", 300_000.seconds) %>
```
anthotsang: should this be added to .env.example and app.json?

hanchang: Yes, good call!

PR #561, opened 2018-11-08
Add Submenu Tabs to Coin Detail Pages

anthotsang:
https://app.asana.com/0/683768070860536/890355139547882

Includes:
https://github.com/coinfi/coinfi/pull/548

PR #567, opened 2018-11-09
Remake coin details page

anthotsang:
https://app.asana.com/0/683768070860536/866838733012110

Includes:
https://github.com/coinfi/coinfi/pull/548
https://github.com/coinfi/coinfi/pull/561


app/controllers/coins_controller.rb
```diff
@@ -83,6 +75,16 @@ def set_coin
...
   end
 
+  def get_token(dashboard_id)
+    payload = {
+      resource: { dashboard: dashboard_id },
+      params: {
+        "coin_key" => @coin.coin_key
+      }
+    }
+    JWT.encode payload, ENV['METABASE_SECRET_KEY']
```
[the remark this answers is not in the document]
anthotsang: oops that method isn't used anymore. thanks for pointing it out.

app/models/coin.rb
```diff
@@ -211,6 +226,84 @@ def sparkline
     end
   end
 
+  def token_metrics_data(metric_type)
+    return nil unless has_token_metrics?
+
+    url = "#{ENV.fetch('COINFI_POSTGREST_URL')}/metrics_chart_view?coin_key=eq.#{coin_key}&metric_type=eq.#{metric_type}&select=percentage,date"
+    response = HTTParty.get(url)
+    JSON.parse(response.body)
```
hanchang: Does it make more sense to cache this method's return value rather than all of the individual accessor methods?

Basically the accessor methods (i.e. `exchange_supply_data`) just calls this master method and retrieves a hash key from it, so by consolidating the cache here it reduces complexity and also prevents multiple calls to this endpoint.

app/javascript/App/images/lightbulb.svg
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
hanchang: Where is this file used?

If it is being used, we should consider replacing it with a FontAwesome icon instead as well.

app/javascript/App/bundles/CoinShow/index.js
```diff
@@ -1,369 +1,979 @@
...
 
-  watchlistHandler(coin) {
-    if (window.location === `/coins/${coin.get('name')}`) {
-      window.location = `/coins/${coin
-        .get('name')
-        .replace(/ /, '-')
-        .toLowerCase()}`
+    const { tokenMetrics } = props
+    const hasTokenMetrics = !_.isEmpty(tokenMetrics)
+    const hashTag = _.get(props, ['location', 'hash'], '').slice(1) // remove prepended octothorpe
```
hanchang: I'm very happy that our codebase has the word "octothorpe" in it now :octocat: 🐙

app/models/coin.rb
```diff
@@ -211,6 +226,68 @@ def sparkline
...
+      JSON.parse(response.body)
+    end
+  end
+
+  def token_metrics_metadata(metric_type)
+    return nil unless has_token_metrics?
+
+    Rails.cache.fetch("coins/#{id}/#{metric_type}_metadata", expires_in: 1.day, race_condition_ttl: 10.seconds) do
+      url = "#{ENV.fetch('COINFI_POSTGREST_URL')}/#{metric_type}_metrics_view?coin_key=eq.#{coin_key}"
+      headers = { "Accept": "application/vnd.pgrst.object+json" }
```
hanchang: Just curious why this header is here but not in the `token_metrics_data` method?

app/models/coin.rb
```diff
@@ -211,6 +226,68 @@ def sparkline
     end
   end
 
+  def token_metrics_data(metric_type)
+    return nil unless has_token_metrics?
+
+    Rails.cache.fetch("coins/#{id}/#{metric_type}", expires_in: 1.day, race_condition_ttl: 10.seconds) do
```
hanchang: Might also be worthwhile from a performance perspective to assign this result to an instance variable so that each call to individual metric methods (i.e. `exchange_supply_data`, `token_retention_rate_data`, etc.) doesn't need to go over the wire to fetch from Redis:

```
@token_metrics_data ||= Rails.cache.fetch...
```

app/models/coin.rb
```diff
@@ -211,6 +226,68 @@ def sparkline
...
+      url = "#{ENV.fetch('COINFI_POSTGREST_URL')}/metrics_chart_view?coin_key=eq.#{coin_key}&metric_type=eq.#{metric_type}&select=percentage,date"
+      response = HTTParty.get(url)
+      JSON.parse(response.body)
+    end
+  end
+
+  def token_metrics_metadata(metric_type)
+    return nil unless has_token_metrics?
+
+    Rails.cache.fetch("coins/#{id}/#{metric_type}_metadata", expires_in: 1.day, race_condition_ttl: 10.seconds) do
```
hanchang: Same here as above

app/models/coin.rb
hanchang: Gotcha, cool thanks - if you could include a link to the docs mentioning this feature that'd be super helpful!

PR #581, opened 2018-11-14
Setup redis followers

anthotsang:
https://app.asana.com/0/683768070860536/909799426526781


app.json
```diff
@@ -2,7 +2,7 @@
   "name": "coinfi-rails",
   "scripts": {
     "postdeploy":
-      "pg_dump $STAGING_DATABASE_URL | psql $DATABASE_URL && bundle exec rake db:migrate"
+      "pg_dump $STAGING_DATABASE_URL | psql $DATABASE_URL && bundle exec rake db:migrate; bundle exec rake redis:copy_prod"
```
hanchang: Hm, should this be `redis:make_slave` instead based on our Slack conversation?

lib/tasks/redis.rake
```diff
@@ -0,0 +1,22 @@
+require "redis"
+
+namespace :redis do
+  desc "Make redis a slave of production redis"
+  task :make_slave => :environment do
+    redis = Redis.new(url: ENV.fetch('REDIS_URL'), :driver => :hiredis)
+    redis.slaveof(ENV.fetch('PRODUCTION_REDIS_HOST'), ENV.fetch('PRODUCTION_REDIS_PORT'))
```
hanchang: I think you can parse the `REDIS_URL` variable to get the constituent parts to keep the env vars DRY.

```
prod_redis_uri = URI.parse(ENV["PRODUCTION_REDIS_URL"])

redis.slaveof(ENV.fetch(uri.host), ENV.fetch(uri.port))
redis.config('set', 'masterauth', ENV.fetch(uri.password))
```

lib/tasks/redis.rake
```diff
@@ -0,0 +1,22 @@
+require "redis"
+
+namespace :redis do
+  desc "Make redis a slave of production redis"
+  task :make_slave => :environment do
+    redis = Redis.new(url: ENV.fetch('REDIS_URL'), :driver => :hiredis)
```
hanchang: Hm, we may need to install the `hiredis` gem in order to use the driver, but last time I checked `hiredis` isn't supported on Heroku.

There's a buildpack to add it but that might be overkill:
https://elements.heroku.com/buildpacks/keiko713/heroku-buildpack-hiredis

I'd suggest holding off and seeing what kind of performance impact there is; if it's bad then we can consider adding it in.

lib/tasks/redis.rake
```diff
@@ -0,0 +1,22 @@
...
+    redis.slaveof(ENV.fetch('PRODUCTION_REDIS_HOST'), ENV.fetch('PRODUCTION_REDIS_PORT'))
+    redis.config('set', 'masterauth', ENV.fetch('PRODUCTION_REDIS_PASSWORD'))
+    redis.config('set', 'slave-read-only', 'no')
```
hanchang: I'm not entirely sure if we want the slave to be writable because theoretically they should all just sync from the primary, since only the production instance will retrieve data from CMC.

lib/tasks/redis.rake
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  desc "One-time copy of production redis"
+  task :copy_prod => :environment do
+    redisSrc = Redis.new :url => ENV.fetch('PRODUCTION_REDIS_URL')
+    redisDest = Redis.new :url => ENV.fetch('REDIS_URL')
+
```
hanchang: Might want to do a check to make sure `PRODUCTION_REDIS_URL != REDIS_URL` just in case...

lib/tasks/redis.rake
hanchang: My bad, looks like things have changed and `hiredis` works on Heroku now!

app.json
anthotsang: Ah sorry, I've been leaving comments in the asana ticket. Heroku redis doesn't allow slaveof

lib/tasks/redis.rake
anthotsang: From the documentation, it allows you to write locally and then it'll just be overwritten upon update. This makes it so that rails doesn't throw an error when it wants to write to redis.

> You may wonder why it is possible to revert the read-only setting and have slave instances that can be targeted by write operations. While those writes will be discarded if the slave and the master resynchronize or if the slave is restarted, there are a few legitimate use case for storing ephemeral data in writable slaves.

> Also note that since Redis 4.0 slave writes are only local, and are not propagated to sub-slaves attached to the instance. Sub slaves instead will always receive the replication stream identical to the one sent by the top-level master to the intermediate slaves. So for example in the following setup:

lib/tasks/redis.rake
anthotsang: Sounds good

lib/tasks/redis.rake
```diff
@@ -0,0 +1,29 @@
+require "redis"
+
+namespace :redis do
+  desc "Make redis a slave of production redis"
+  task :make_slave => :environment do
+    prod_redis_uri = URI.parse(ENV.fetch('PRODUCTION_REDIS_URL'))
+    next if prod_redis_uri.blank?
```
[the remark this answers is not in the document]
anthotsang: From what I can tell, rake tasks throw an error (warning?) if you `return` early and the appropriate way is to use `next` instead.

lib/tasks/redis.rake
```diff
@@ -0,0 +1,29 @@
...
+    redis.slaveof(prod_redis_uri.host, prod_redis_uri.port)
+    redis.config('set', 'masterauth', prod_redis_uri.password)
+    redis.config('set', 'slave-read-only', 'no')
+  end
+
+  desc "One-time copy of production redis"
+  task :copy_prod => :environment do
+    prod_redis_uri = ENV.fetch('PRODUCTION_REDIS_URL')
+    local_redis_uri = ENV.fetch('REDIS_URL')
+    next if prod_redis_uri.blank? || prod_redis_uri == local_redis_uri
```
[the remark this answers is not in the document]
anthotsang: From what I can tell, rake tasks throw an error (warning?) if you return early and the appropriate way is to use next instead.

lib/tasks/redis.rake
```diff
@@ -0,0 +1,29 @@
...
+
+    redis = Redis.new(url: ENV.fetch('REDIS_URL'), :driver => :hiredis)
+    redis.slaveof(prod_redis_uri.host, prod_redis_uri.port)
+    redis.config('set', 'masterauth', prod_redis_uri.password)
+    redis.config('set', 'slave-read-only', 'no')
+  end
+
+  desc "One-time copy of production redis"
+  task :copy_prod => :environment do
+    prod_redis_uri = ENV.fetch('PRODUCTION_REDIS_URL')
```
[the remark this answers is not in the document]
anthotsang: I think we need to set `PRODUCTION_REDIS_URL: true` in app.json to make sure it's copied to the review apps.

As far as I can tell, we can just set `PRODUCTION_REDIS_URL` to nothing on production and it'll `fetch` without error and still return `nil`? But that may just be a quirk with how dotenv works vs how heroku handles environmental variables.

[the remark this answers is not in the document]
anthotsang: I guess we can still just not use fetch anyways though.

lib/tasks/redis.rake
```diff
@@ -0,0 +1,30 @@
...
+    prod_redis_uri = URI.parse(raw_prod_redis_uri)
+    redis = Redis.new(url: ENV.fetch('REDIS_URL'), :driver => :hiredis)
+    redis.slaveof(prod_redis_uri.host, prod_redis_uri.port)
+    redis.config('set', 'masterauth', prod_redis_uri.password)
+    redis.config('set', 'slave-read-only', 'no')
+  end
+
+  desc "One-time copy of production redis"
+  task :copy_prod => :environment do
+    prod_redis_uri = ENV['PRODUCTION_REDIS_URL']
```
hanchang: This might need to also be `ENV.fetch('PRODUCTION_REDIS_URL')`, otherwise `prod_redis_uri` is `nil` and we'd call `redisSrc = Redis.new :url => nil` which is a valid line and returns the localhost version of Redis (no idea what would happen there since there isn't a Redis installation on the Heroku instance)

hanchang: My bad, this is handled by line 20!

PR #584, opened 2018-11-15
Fix subscription watchlist out of sync #911685984649517

jessehon:
https://app.asana.com/0/904189350705878/911685984649517

### DIFF
--- a/lib/tasks/data_migrations/11_sync_watchlist_items_with_signals_telegram_subscriptions.rake
+++ b/lib/tasks/data_migrations/11_sync_watchlist_items_with_signals_telegram_subscriptions.rake
@@ -1,0 +1,44 @@
+namespace :data_migrations do
+  desc <<~DESC
+      Sync the WatchlistItems so that they reflect the current state of SignalsTelegramSubscriptions
+    DESC
+  task :sync_watchlist_items_with_signals_telegram_subscriptions => :environment do
+    ActiveRecord::Base.transaction do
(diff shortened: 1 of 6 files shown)


lib/tasks/data_migrations/11_sync_watchlist_items_with_signals_telegram_subscriptions.rake
```diff
@@ -0,0 +1,43 @@
...
+  task :sync_watchlist_items_with_signals_telegram_subscriptions => :environment do
+    ActiveRecord::Base.transaction do
+      SignalsTelegramUser.find_each do |signals_telegram_user|
+        watchlist = signals_telegram_user.user.watchlist
+        subscribed_coins = signals_telegram_user.subscribed_coins
+        watchlist_coins = watchlist.coins
+
+        watchlist_coins_to_add = subscribed_coins.select do |subscribed_coin|
+          watched = watchlist_coins.any? { |watchlist_coin| watchlist_coin.id == subscribed_coin.id }
+          next !watched
```
hanchang: Should this be `next if !watched` ?

hanchang: Thanks for the offline explanation!

jessehon: For future reference https://stackoverflow.com/a/11026130

PR #585, opened 2018-11-16
Switch from Heroku Scheduler to Sidekiq Scheduler

anthotsang:
https://app.asana.com/0/683768070860536/793009846303982


config/sidekiq_production.yml
```diff
@@ -0,0 +1,18 @@
+SparklineScheduler:               # rake cache_refresh:sparkline
+  cron: '0 0 19 * * * Etc/UTC'
+
+CoinSlugsScheduler:               # rake generate_coin_slugs
+  cron: '0 0 1 * * * Etc/UTC'
+
+SitemapScheduler:                 # rake sitemap:refresh
+  cron: '0 0 0 * * * Etc/UTC'
+
```
hanchang: Hm, might be worth consolidating around a standardized naming and directory layout convention; here's my proposal:

Class names should be descriptive and start with a verb, i.e. `RefreshSparkline`. 

All Scheduler jobs are located in the `lib/scheduled_jobs` directory. As a result, there's no need to append `Scheduler` to the class name.

Lemme know your thoughts on this!

config/sidekiq_production.yml
```diff
@@ -0,0 +1,18 @@
...
+staked_cofi_perform_actions:      # rake staked_cofi:perform_actions
+  cron: '0 30 * * * *'
+  class: StakedCofiScheduler
+  args: 'perform_actions'
+
+staked_cofi_refresh_transactions: # rake staked_cofi:refresh_transactions
+  cron: '0 0 * * * *'
+  lass: StakedCofiScheduler
```
hanchang: I think we're missing the `c` in `class` here.

config/sidekiq_production.yml
```diff
@@ -0,0 +1,18 @@
...
+  cron: '0 0 19 * * * Etc/UTC'
+
+CoinSlugsScheduler:               # rake generate_coin_slugs
+  cron: '0 0 1 * * * Etc/UTC'
+
+SitemapScheduler:                 # rake sitemap:refresh
+  cron: '0 0 0 * * * Etc/UTC'
+
+staked_cofi_perform_actions:      # rake staked_cofi:perform_actions
+  cron: '0 30 * * * *'
```
hanchang: Thanks for including the timezone in the previous cron declarations; let's stick with that here as well for consistency.

config/sidekiq_production.yml
```diff
@@ -0,0 +1,18 @@
...
+  class: StakedCofiScheduler
+  args: 'perform_actions'
+
+staked_cofi_refresh_transactions: # rake staked_cofi:refresh_transactions
+  cron: '0 0 * * * *'
```
hanchang: Thanks for including the timezone in the previous cron declarations; let's stick with that here as well for consistency.

lib/staked_cofi_scheduler.rb
```diff
@@ -1,7 +1,27 @@
-# Creates or updates StakedCofiTransaction items for every COFI transaction found on EtherScan
-namespace :staked_cofi do
-  desc "Refresh staked cofi transactions by querying etherscan"
-  task :refresh_transactions => :environment do
+require 'sidekiq-scheduler'
+class StakedCofiScheduler
```
hanchang: It's probably worthwhile to split these into two separate jobs and files.

config/sidekiq_production.yml
anthotsang: I didn't include it for any tasks that wouldn't be affected by it (no hour or greater time unit declarations), but I can include it for all just for consistency?

config/sidekiq_production.yml
hanchang: Basically the idea behind this standardization is to make it easier to tell at a glance what each scheduler job class does.

For example, if a dev sees `CoinSlugsScheduler` in the `sidekiq_production.yml` file, it's not immediately obvious that the job is to generate coin slugs, whereas `GenerateCoinSlugs` makes it more clear without having to look at the source.

anthotsang: Yeah sounds good. I initially put them all there because only the root lib folder was being autoloaded. But I guess it makes more sense to just add another folder to autoloading than to avoid putting them all into a folder.

Procfile
```diff
@@ -1 +1,2 @@
 web: bundle exec puma -C config/puma.rb
+worker: bundle exec sidekiq -e production -C config/sidekiq.yml
```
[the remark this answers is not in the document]
anthotsang: [Default concurrency has been reduced to 10 for sidekiq.](https://github.com/mperham/sidekiq/commit/8ff96ae0b0358dc273d19ae1f8474f6ff4fd2b64)

Changing command to `env MALLOC_ARENA_MAX=2 bundle exec sidekiq -e production -C config/sidekiq.yml` based on [this](https://www.mikeperham.com/2018/04/25/taming-rails-memory-bloat/)

config/initializers/sidekiq.rb
```diff
@@ -0,0 +1,26 @@
+require 'sidekiq'
+require 'sidekiq-scheduler'
+
+is_production = (ENV['IS_PRODUCTION'] || "false").downcase == 'true'
```
hanchang: 💪 `downcase` ftw, just in case...

config/initializers/sidekiq.rb
```diff
@@ -0,0 +1,26 @@
...
+redis_url = ENV.fetch('REDISCLOUD_URL') if is_production
```
hanchang: Should we just wrap this entire thing in a `if is_production` block?

PR #587, opened 2018-11-19
Retrieve and store CoinMarketCap Pro Exchange Breakdown Data

anthotsang:
https://app.asana.com/0/902343675126966/892701094146974


app/services/coin_market_cap_service.rb
```diff
@@ -1,40 +1,168 @@
 require_relative '../../lib/tasks/batch_process'
 
 class CoinMarketCapService
```
hanchang: Totally my bad, but basically this file is historically named incorrectly; it should be called `NorthPoleProxyService` or something since it only retrieves information from the NorthPole CMC proxy, not CMC directly. There's no need to rename it though since we're planning on deleting this file soon anyways.

Instead, can we create a new `CoinMarketCapProService` file that actually contains only logic pertaining to the new CMC Pro API and revert this file back to the original?

This will also make it easier to perform the code review as well as delete this file when it's no longer needed. Thanks!

config/sidekiq.yml
```diff
@@ -0,0 +1,70 @@
+:schedule:
```
hanchang: Just curious if this file is supposed to be separate from `config/sidekiq_production.yml`?

https://github.com/coinfi/coinfi/pull/585/files#diff-3924c9053e4f667358528b94fd4df107

config/sidekiq.yml
```diff
@@ -0,0 +1,70 @@
...
+  cmc_update_prices1_scheduler:
```
hanchang: It might make more sense to just have the scheduler run every minute but have the logic within the job handle what to update (via something like modulo division).

This way it reduces the amount of jobs in the queue, otherwise things can get a little crazy in Redis.

config/sidekiq.yml
anthotsang: No, I made this before making the change in the other pull request. Wasn't sure if the other change would work properly or not so I hadn't updated this file yet.

app/services/coin_market_cap_service.rb
anthotsang: sure

lib/scheduled_jobs/update_cmc_prices.rb
```diff
@@ -0,0 +1,29 @@
+require 'sidekiq-scheduler'
+class UpdateCmcPrices
+  include Sidekiq::Worker
+
+  def perform(metadata)
+    minutes = Time.at(metadata["scheduled_at"]).min
```
hanchang: Nice, very elegant!

lib/tasks/coinmarketcap.rake
```diff
@@ -2,10 +2,10 @@ require 'httparty'
 
 namespace :coinmarketcap do
   namespace :cronjobs do
-    # Direct ticker access
-    desc "10min ticker price update"
+    # Direct suppy access
+    desc "Daily suppy update"
     task :ticker_update => :environment do
-      CoinMarketCapService.new.ticker_update
+      CoinMarketCapService.new.supply_update
```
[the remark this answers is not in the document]
anthotsang: I'm reverting back to the original code since this is for the current ticker update. New stuff doesn't touch this file, this is still just the northpole rake task.

lib/tasks/coinmarketcap.rake
```diff
@@ -2,10 +2,10 @@ require 'httparty'
 
 namespace :coinmarketcap do
   namespace :cronjobs do
-    # Direct ticker access
-    desc "10min ticker price update"
+    # Direct suppy access
+    desc "Daily suppy update"
     task :ticker_update => :environment do
```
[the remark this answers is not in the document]
anthotsang: I'm reverting back to the original code since this is for the current ticker update. New stuff doesn't touch this file, this is still just the northpole rake task.

PR #595, opened 2018-11-26
Allow CMC Pro Volume 24 to be null

anthotsang:
https://app.asana.com/0/902343675126966/921512712647969

### DIFF
--- a/app/services/coin_market_cap_pro/update_market_pairs_service.rb
+++ b/app/services/coin_market_cap_pro/update_market_pairs_service.rb
@@ -38,6 +38,7 @@ module CoinMarketCapPro
       # Grabbing data from snapshot cache
       snapshot = Rails.cache.read("#{identifier}:snapshot")
       base_volume24 = snapshot[:volume24h] unless snapshot.blank?
+      has_base_volume24 = !base_volume24.blank? && base_volume24 != 0
 
       raw_market_pairs = data.dig("market_pairs")
       market_pairs = raw_market_pairs.map do |pair|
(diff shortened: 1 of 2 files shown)


app/services/coin_market_cap_pro/update_market_pairs_service.rb
```diff
@@ -38,6 +38,7 @@ def perform_update_pairs(identifier, data)
       # Grabbing data from snapshot cache
       snapshot = Rails.cache.read("#{identifier}:snapshot")
       base_volume24 = snapshot[:volume24h] unless snapshot.blank?
+      has_base_volume24 = !base_volume24.blank? && base_volume24 != 0
```
hanchang: Might be easier just to use `base_volume24.present?`

app/services/coin_market_cap_pro/update_market_pairs_service.rb
```diff
@@ -50,7 +51,7 @@ def perform_update_pairs(identifier, data)
           :pair => pair["market_pair"],
           :price => quote["price"],
           :volume24h => volume24,
-          :volume_percentage => (volume24 / base_volume24 unless base_volume24.blank?),
+          :volume_percentage => (volume24 / base_volume24 if has_base_volume24),
```
hanchang: Maybe an if/else might be better here? Otherwise is `volume_percentage` going to be null?

i.e. `:volume_percentage => (if has_base_volume24 then volume24 / base_volume24 else 0 end),

PR #597, opened 2018-11-27
Set supply fields to optional

anthotsang:
https://app.asana.com/0/902343675126966/923519597998211

### DIFF
--- a/app/services/coin_market_cap_pro/update_ticker_service.rb
+++ b/app/services/coin_market_cap_pro/update_ticker_service.rb
@@ -46,11 +46,10 @@ module CoinMarketCapPro
     end
 
     def has_missing_data(data)
-      quote = data.dig("quote", currency)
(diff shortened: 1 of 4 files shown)


app/services/coin_market_cap_pro/update_ticker_service.rb
```diff
@@ -46,11 +46,10 @@ def log_missing_data
     end
 
     def has_missing_data(data)
-      quote = data.dig("quote", currency)
-      if quote["price"].blank? || quote["market_cap"].blank? ||
-        quote["percent_change_1h"].blank? || quote["percent_change_24h"].blank? ||
-        quote["percent_change_7d"].blank? || data["total_supply"].blank? ||
-        data['circulating_supply'].blank? then
+      quote = data.dig('quote', currency)
+      if quote['price'].blank? || quote['market_cap'].blank? ||
```
hanchang: Just realized we can just return this statement directly.

PR #600, opened 2018-11-27
Ensure watchlist and signal subscriptions are in sync when adding via newsfeed #923530073585132

jessehon:
https://app.asana.com/0/904189350705878/923530073585132

### DIFF
--- a/app/controllers/api/watchlist/coins_controller.rb
+++ b/app/controllers/api/watchlist/coins_controller.rb
@@ -8,19 +8,18 @@ class Api::Watchlist::CoinsController < ApiController
 
   def create
     @coin = Coin.find(params[:id])
-    if @watchlist.coins.find_by_id(@coin.id)
+    if @watchlist.coins.where(id: @coin.id).exists?
       respond_warning "Coin already added"
     else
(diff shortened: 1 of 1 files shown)


app/controllers/api/watchlist/coins_controller.rb
```diff
@@ -8,19 +8,18 @@ def index
...
+      WatchCoinService.call(user: current_user, coin: @coin)
       respond_success "Coin added to watchlist"
     end
   end
 
   def destroy
-    if @item = @watchlist.items.find_by_coin_id(params[:id])
-      @item.destroy
-      respond_success({ id: @item.coin_id }, "Coin removed from watchlist")
+    if @coin = @watchlist.coins.find_by_id(params[:id])
```
hanchang: I think this can just be `@watchlist.coins.find(params[:id])`, no need for `find_by_id`

jessehon: `.find` behaves a bit differently actually. Instead of a boolean result, it would throw a `ActiveRecord::NotFound` error if the record is not found

PR #603, opened 2018-11-28
Add Registration CTA to bottom of all CoinFi News articles

hanchang:
https://app.asana.com/0/902343675126966/921631352220749

`alignLeft` is a janky solution I just threw together; the default is to be center aligned, so `alignLeft` is an optional param specifically for Twitter to make the layout look reasonable.

This PR also includes Heap instrumentation.

screen shot 2018-11-28 at 11 56 23 am
screen shot 2018-11-28 at 12 58 54 pm

### DIFF
--- a/app/javascript/App/bundles/NewsfeedPage/NewsBody.tsx
+++ b/app/javascript/App/bundles/NewsfeedPage/NewsBody.tsx
@@ -8,6 +8,7 @@ import Icon from '~/bundles/common/components/Icon'
 import localAPI from '../common/utils/localAPI'
 
 import TwitterBody from './TwitterBody'
+import CallToAction from './CallToAction'
 import LoadingIndicator from '../common/components/LoadingIndicator'
 import {
   getDomainType,
@@ -165,6 +166,9 @@ export default class NewsBody extends React.Component<Props, State> {
             }}
           </RailsConsumer>
         </div>
+
+        <br />
+        <CallToAction />
       </div>
     )
   }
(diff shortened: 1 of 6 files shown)


app/javascript/App/bundles/NewsfeedPage/TwitterBody.tsx
```diff
@@ -23,23 +24,26 @@ export default class NewsBody extends React.Component<Props, {}> {
...
-            {categories.map((category, index) => (
-              <div key={index} className="tag-alt">
-                {category.name}
-              </div>
-            ))}
-          </div>
-        )}
-        <Tweet tweetId={tweetId} />
-      </div>
+      <>
```
anthotsang: this doesn't look like it's necessary

app/javascript/App/bundles/NewsfeedPage/NewsBody.tsx
```diff
@@ -165,6 +166,9 @@ export default class NewsBody extends React.Component<Props, State> {
             }}
           </RailsConsumer>
         </div>
+
+        <br />
+        <CallToAction />
```
anthotsang: should this only be here if not logged in?

app/javascript/App/bundles/NewsfeedPage/TwitterBody.tsx
hanchang: Great catch, thanks - removing...

app/javascript/App/bundles/NewsfeedPage/NewsBody.tsx
hanchang: Was hoping to not have to prop drill `loggedIn` but it's a better UX so will just bite the bullet, thanks.

PR #610, opened 2018-11-29
Fix toplist

anthotsang:
https://app.asana.com/0/902343675126966/926961739608610


app/helpers/coin_list_helper.rb
```diff
@@ -0,0 +1,25 @@
+module CoinListHelper
+  def toplist_coins
+    distribute_reads(max_lag: ApplicationController::MAX_ACCEPTABLE_REPLICATION_LAG, lag_failover: true) do
+      serialized_coins = Rails.cache.fetch("coins/toplist", expires_in: 1.hour) do
```
hanchang: The cache key stays the same but the cache value will change (serialized values vs non-serialized values) so we need to remember to clear Rails cache when deploying - please note in the deployment docs, thanks!

PR #611, opened 2018-11-29
Add api route to show coin for telegram bot #926978408683030

jessehon:
https://app.asana.com/0/904189350705878/926978408683030


test/integration/api/signals_telegram_bot/signals_telegram_users/render_test.rb
```diff
@@ -2,7 +2,7 @@
 require 'test_helper'
 require_relative './base_test'
```
hanchang: Think this needs to be `../base_test` due to the file relocation

PR #613, opened 2018-11-30
Enable CMC market pair updating.

anthotsang:
https://app.asana.com/0/902343675126966/928801964945187

### DIFF
--- a/config/sidekiq_production.yml
+++ b/config/sidekiq_production.yml
@@ -23,41 +23,41 @@ update_cmc_market_pairs1:
   args: 1
 
 update_cmc_market_pairs2:
-  cron: '30 */5 * * * *' # Runs once every 5 minutes
+  cron: '31 */5 * * * *' # Runs once every 5 minutes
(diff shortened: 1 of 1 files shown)


config/sidekiq_production.yml
```diff
@@ -17,47 +17,47 @@ UpdateCmcPrices:
...
-#   cron: '30 0 */4 * * *' # Runs once every 4 hours
-#   class: UpdateCmcMarketPairs
-#   args: 8
-
-# update_cmc_market_pairs9:
-#   cron: '30 0 */5 * * *' # Runs once every 5 hours
-#   class: UpdateCmcMarketPairs
-#   args: 9
+update_cmc_market_pairs1:
+  cron: '30 */2 * * * *' # Runs once every other minute
```
hanchang: Hm, wondering if it's worth it to space out the start times by seconds?

Just a bit worried that at the `:30` second mark there will be too many jobs queued simultaneously and it's actually unclear which ones will take precedence.

PR #618, opened 2018-12-05
Remove references to "reservation" in emails to match "access language" on landing page

jlei523:


hanchang:
Looks good @jlei523 - could you just start a fresh branch off of develop with these changes and fix the broken test?

PR #619, opened 2018-12-05
Daily abnormal token movements signal #921575776021947

jessehon:
https://app.asana.com/0/904189350705878/913441027569130


app/models/trading_signal.rb
```diff
@@ -6,4 +6,14 @@ class TradingSignal < ApplicationRecord
   validates :external_id, uniqueness: true
   validates :trading_signal_trigger_id, presence: { unless: :trading_signal_trigger_external_id? }
   validates :trading_signal_trigger_external_id, presence: { unless: :trading_signal_trigger_id? }
+
+  def as_dto
```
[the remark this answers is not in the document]
jessehon: Data transfer object

lib/scheduled_jobs/daily_token_exchange_transactions_trading_signal_worker.rb
```diff
@@ -0,0 +1,8 @@
+class DailyTokenExchangeTransactionsTradingSignalWorker
+  include Sidekiq::Worker
+
+  def perform
+    trading_signal = TradingSignals::DailyTokenExchangeTransactions.build()
```
hanchang: I think no parens are needed, writing too much JavaScript 😛

.env.example
```diff
@@ -28,6 +28,9 @@ ETL_DB_NAME=
 ETL_DB_PASSWORD=
 ETL_DB_PORT=5432
 ETL_DB_USER=
+GOOGLE_CREDENTIALS_JSON=
```
hanchang: @jessehon Just wanted to check - these env vars don't need to be in `app.json` because they're only loaded in the worker which doesn't matter on review apps, correct?

jessehon: Nope, just forgot to add them haha

PR #622, opened 2018-12-07
Fix market pair update service

anthotsang:
https://app.asana.com/0/902343675126966/937965094554593


app/services/coin_market_cap_pro/update_market_pairs_service.rb
```diff
@@ -12,10 +11,32 @@ def initialize(start: 0, limit: 20) # 0-indexed
 
     def call
       coins = Coin.top(@limit).offset(@start)
-      batch_process(coins) do |coin|
-        update_coin_pairs(coin)
+      item_name = coins.first.class.name
```
hanchang: Can probably just be `Coin.name`

PR #623, opened 2018-12-07
Turn off SSR for coin details

anthotsang:
https://app.asana.com/0/902343675126966/941872991961444


app/javascript/App/bundles/CoinShow/InfoBar.tsx
```diff
@@ -112,6 +112,7 @@ class InfoBar extends React.Component<Props, {}> {
       market_cap,
       change1h,
       change24h,
+      volume24,
```
hanchang: Hm I think we should be consistent and call this `volume24h`

anthotsang: Yeah there's a consistency issue there that should be resolved.

app/models/coin.rb
```diff
@@ -184,6 +184,21 @@ def market_info(market_data = nil)
     data
   end
 
+  def token_metrics
```
hanchang: Hm, from my understanding, this method will make 10 GET requests on a cold cache, will that be a performance problem?

anthotsang: Hasn't seemed to be an issue. We should definitely look to improving how we fetch from cache as the amount of data we get comes from cache. But, compared to the initial 10 remote calls, I think it's been pretty trivial.

We can try leveraging read_multi, but we can't just replace it directly since we'll have to also check/fetch any missing keys afterwards.

PR #628, opened 2018-12-10
Improve populate token decimals task

jessehon:


lib/tasks/data_migrations/12_populate_token_decimals_on_coins.rake
```diff
@@ -8,6 +8,7 @@ namespace :data_migrations do
 
     batch_process(coins_to_update) do |coin|
       PopulateTokenDecimalsForCoinWorker.new.perform(coin)
+      sleep 0.5
```
hanchang: I'm guessing this is to stay under the Etherscan rate limit?

hanchang: Sorry I meant Ethplorer

jessehon: Yup

PR #546, opened 2018-11-05
New Homepage Setup

anthotsang:
https://app.asana.com/0/683768070860536/885096854820729

PR #641, opened 2018-12-14
Store total market cap/volume

anthotsang:
https://app.asana.com/0/902343675126966/932612706107745


.env.example
```diff
@@ -34,12 +34,13 @@ GOOGLE_PROJECT_ID=coinfi-dev
 GOOGLE_PUBSUB_TRADING_SIGNALS_TOPIC_NAME=projects/coinfi-dev/topics/trading_signals
 HEALTHCHECK_DAILY_PRICES=
 HEALTHCHECK_HOURLY_PRICES=
+HEALTHCHECK_MARKET_METRICS=
+HEALTHCHECK_MARKET_PAIRS=
```
hanchang: Thanks for alphabetizing! 🔤

app/services/coin_market_cap_pro/update_market_metrics_service.rb
```diff
@@ -0,0 +1,43 @@
...
+    private
+
+    def update_metrics
+      data = load_cmc_data
+      perform_update_metrics(data) if data.present?
+    end
+
+    def perform_update_metrics(data)
+      metrics = data.dig('quote', 'USD')
+      MarketMetric.create(
```
hanchang: Should we do a `create!` instead to throw an exception if anything goes awry?

Otherwise we could still fire the Market Metrics healthcheck even if the insertion was a failure...

anthotsang:
Re: `distributed_reads`, I assume we can't wrap scopes directly, so I guess this I will make a note of it when calling this table for #546

app/services/coin_market_cap_pro/update_market_metrics_service.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    def perform_update_metrics(data)
+      metrics = data.dig('quote', 'USD')
+      begin
+        MarketMetric.create!(
+          total_market_cap: metrics["total_market_cap"],
+          total_volume_24h: metrics["total_volume_24h"],
+          timestamp: metrics["last_updated"],
+        )
+      rescue StandardError => e
```
hanchang: No need to change but the Rubyesque thing is to just do a `rescue => error` since `StandardError` is the default.

PR #666, opened 2019-01-15
Add currency dropdown to home and coin index

anthotsang:
https://app.asana.com/0/902343675126966/980409871436053


app/helpers/currency_helper.rb
```diff
@@ -0,0 +1,9 @@
+module CurrencyHelper
```
hanchang: I'd personally prefer removing this class and just being explicit in the controllers but not a big deal either way, will leave it up to you.

PR #675, opened 2019-01-31
ETH Signals

anthotsang:
https://app.asana.com/0/902343675126966/1003124458420333


app/controllers/coins_controller.rb
```diff
@@ -38,6 +38,11 @@ def show
         @top_coins_data = toplist_coins
         @watched_coins_data = watchlist_coins if current_user
       end
+
+      if is_ethereum?(@coin)
```
hanchang: Why is the parameter being passed in an instance variable?

PR #681, opened 2019-02-07
Update exchange data service

anthotsang:
https://app.asana.com/0/902343675126966/1108365461183592


app/services/coin_market_cap_pro/coin_market_cap_pro_helpers.rb
```diff
@@ -11,5 +11,41 @@ def get_error_message(contents)
...
+    end
+
+    def log_or_ping_on_missing_data(missing_data, healthcheck_url = nil)
+      if missing_data.empty?
+        Net::HTTP.get(URI.parse(healthcheck_url)) unless healthcheck_url.blank?
+      else
+        if healthcheck_url.present?
+          Net::HTTP.post(URI.parse("#{healthcheck_url}/fail"), missing_data.to_json)
+        else
+          pp missing_data
```
hanchang: Will pretty printing log to Papertrail?

app/services/coin_market_cap_pro/coin_market_cap_pro_helpers.rb
```diff
@@ -11,5 +11,41 @@ def get_error_message(contents)
...
+
+      # ping health check if api error
+      json_response_code = get_json_response_code(contents)
+
+      if response.success? && json_response_code == 0 then
+        return contents['data']
+      end
+
+      error_message = "ERROR HTTP(#{response.code}) JSON(#{json_response_code}): #{get_error_message(contents)}"
+      if healthcheck_url.present?
```
hanchang: Nice catch just to make sure 👍

app/services/coin_market_cap_pro/coin_market_cap_pro_helpers.rb
```diff
@@ -11,5 +11,41 @@ def get_error_message(contents)
...
+        Net::HTTP.post(URI.parse("#{healthcheck_url}/fail"), error_message)
+      else
+        puts error_message
```
hanchang: Does it make sense to throw a Rollbar error here in case `healthcheck_url` is not present? Genuinely asking.

app/services/coin_market_cap_pro/coin_market_cap_pro_helpers.rb
hanchang: (Same question regarding Rollbar here as well)

app/services/coin_market_cap_pro/update_exchanges_service.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    def extract_ids_from_mapping(mapping)
+      mapping.map { |exchange| exchange["id"] }
+    end
+
+    def fetch_exchange_mapping_data(active: true)
+      api_url = "https://pro-api.coinmarketcap.com/v1/exchange/map"
+      query = { :listing_status => active ? 'active' : 'inactive' }
+      headers = get_default_api_headers
+      response = HTTParty.get(api_url, :query => query, :headers => headers)
```
hanchang: Might be worth it to do a try/catch here in case the request fails.

app/services/coin_market_cap_pro/update_exchanges_service.rb
```diff
@@ -0,0 +1,106 @@
...
+      # only ping healthcheck regarding active mapping
+      extract_api_data(response, active ? @healthcheck_url : nil)
+    end
+
+    def fetch_exchange_meta_data(ids)
+      api_url = "https://pro-api.coinmarketcap.com/v1/exchange/info"
+      id_list = ids.respond_to?(:join) ? ids.join(',') : ids
+      query = { :id => id_list }
+      headers = get_default_api_headers
+      response = HTTParty.get(api_url, :query => query, :headers => headers)
```
hanchang: Might be worth it to do a try/catch here in case the request fails.

app/services/coin_market_cap_pro/coin_market_cap_pro_helpers.rb
anthotsang: I made healthchecks optional for helpers in case some things don't use healthchecks.
I think at the service level, we're fetching with `ENV.fetch`, so I assume this would lead to rollbar firing if it was needed but missing?

app/services/coin_market_cap_pro/coin_market_cap_pro_helpers.rb
anthotsang: For papertrail, I think it should since `pp` seems to output to `stdout`. I've converted it anyways just in case.

For rollbar, I think it won't log for `puts` or `pp` since there would have to be an error or you would have to manually log to rollbar from my understanding?

app/services/coin_market_cap_pro/update_exchanges_service.rb
hanchang: @anthotsang Might as well also add the catch here?

anthotsang: Weird. Did everything but this one :man_facepalming: 
Thanks for the catch!

PR #683, opened 2019-02-08
Fix price and volume data.

anthotsang:
https://app.asana.com/0/902343675126966/1108556351143220

### DIFF
--- a/app/services/coin_market_cap_pro/update_market_pairs_service.rb
+++ b/app/services/coin_market_cap_pro/update_market_pairs_service.rb
@@ -79,7 +79,7 @@ module CoinMarketCapPro
         base_symbol = pair.dig("market_pair_base", "currency_symbol")
         coin_is_primary = base_symbol == coin_symbol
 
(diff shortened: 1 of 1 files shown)


app/services/coin_market_cap_pro/update_market_pairs_service.rb
```diff
@@ -59,27 +59,47 @@ def update_coin_pairs(coin)
...
+        pair.dig("quote", currency, "volume_24h") == -1
+      end
+      market_pairs = valid_market_pairs.map do |pair|
         quote = pair.dig("quote", currency)
         volume24h = quote["volume_24h"] || 0
 
+        base_symbol = pair.dig("market_pair_base", "currency_symbol")
+        coin_is_primary = base_symbol == coin_symbol
+
+        volume_percentage = has_base_volume24h ? volume24h / base_volume24h : 0
```
hanchang: Will the result of the division be an `int` or a `float`?

anthotsang: I think it's not a big deal since we display integers, but I will force to float anyways

PR #688, opened 2019-02-19
Bugfix: Apply Reddit/Twitter filters properly for coin news pages

anthotsang:
https://app.asana.com/0/902343675126966/936075974905985

### DIFF
--- a/app/queries/news_items/with_filters.rb
+++ b/app/queries/news_items/with_filters.rb
@@ -1,5 +1,4 @@
 module NewsItems
-  NUMBER_OF_TOP_COINS = 
   class WithFilters
     def self.call(
       relation = NewsItem.all,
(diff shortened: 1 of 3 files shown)


app/queries/news_items/with_filters.rb
```diff
@@ -1,4 +1,5 @@
 module NewsItems
+  NUMBER_OF_TOP_COINS =
```
hanchang: Er what's going on here? Doesn't look like there's a right hand side assignment to the `NUMBER_OF_TOP_COINS` constant, unless I'm missing something?

app/queries/news_items/with_filters.rb
```diff
@@ -12,11 +13,21 @@ def self.call(
...
-      else
+      if feed_sources.present?
         result = result.where(feed_source: feed_sources)
+      else
+        # Based on app/javascript/App/bundles/NewsfeedPage/utils.ts
+        # If all selected coins are top 5 coins (@top_coin_slugs),
+        # then we disable Reddit and Twitter
+        # TODO: Reconcile split back/front-end logic
+        top_coin_slugs = Coin.top(5).pluck(:slug)
+        has_all_top_coins = coins.length <= top_coin_slugs.length &&
```
hanchang: Probably worth refactoring a bit for readability:

```
has_all_top_coins = if coins.present?
  (coins.length <= top_coin_slugs.length) &&
  coins.all? do |coin|
    top_coin_slugs.any? { |top_slug| top_slug == coin.slug }
  end
end
```

There's probably a more elegant way to express this but whatevs.

app/queries/news_items/with_filters.rb
anthotsang: Nope was going to take it out then decided it was unnecessary. Didn't realized that got left there :disappointed:

PR #695, opened 2019-03-01
News voting

anthotsang:
https://app.asana.com/0/902343675126966/1111661820598640
https://app.asana.com/0/902343675126966/1112125718847845


app/controllers/api/news_vote_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    if not news_vote
+      return respond_error 'Could not save vote.'
+    end
+
+    respond_success serialize_vote_summary(news_vote)
+  end
+
+  private
+
+  def serialize_vote_summary(news_vote = nil, summary: nil, vote: nil)
```
hanchang: Just curious if there's a benefit to overriding this method? If not, I would separate this into two different methods, one to handle `news_vote` and the other for `summary`, i.e.

```
def serialize_news_vote(news_vote)
  news_vote.as_json(only: %i[vote], methods: %i[vote_summary])
end

def serialize_news_summary(summary, vote: nil)
  obj = { vote_summary: summary }
  if vote.present?
    obj[:vote] = vote.vote
  end
  obj.as_json
end
```

hanchang: `serialize_news_vote` may not even be needed

app/controllers/webhooks/websubs_controller.rb
```diff
@@ -17,7 +15,7 @@ def superfeedr_ingest
 
     # Rewrite default_news_items cache if any items.
     if items.count > 0
-      get_default_news_items(rewrite_cache: true)
+      NewsServices::get_default_news_items(rewrite_cache: true)
```
hanchang: I like this style better as it's more obvious that this is a call to a service object versus the other calls that route through a helper first which is non-standard.

app/helpers/news_helper.rb
```diff
@@ -38,12 +24,15 @@ def serialize_news_items(news_items)
...
-        news_items = backup_default_news_query if news_items.empty? || news_items.length < DEFAULT_NEWS_LIMIT
-
-        serialize_news_items(news_items)
-      end
+      news_items = NewsServices::RetrieveDefaultNewsItems.call.try(:result)
+      serialize_news_items(news_items) if news_items.present?
     end
   end
+
+  def get_news_items_with_votes(json_news_items)
```
hanchang: Can we just call this directly in the controller when needed instead of routing through the Helper?

app/helpers/news_helper.rb
```diff
@@ -38,12 +24,15 @@ def serialize_news_items(news_items)
 
   def get_default_news_items(rewrite_cache: false)
     Rails.cache.fetch("default_news_items", force: rewrite_cache) do
-      distribute_reads(max_lag: ApplicationController::MAX_ACCEPTABLE_REPLICATION_LAG, lag_failover: true) do
-        news_items = default_news_query
-        news_items = backup_default_news_query if news_items.empty? || news_items.length < DEFAULT_NEWS_LIMIT
-
-        serialize_news_items(news_items)
-      end
+      news_items = NewsServices::RetrieveDefaultNewsItems.call.try(:result)
```
hanchang: Can we just call this directly in the controller when needed instead of routing through the Helper?

app/models/news_vote.rb
```diff
@@ -0,0 +1,104 @@
...
+      .reduce(default_hash) do |hash, item|
+        hash[item.news_item_id] = self.vote_summary(
+          id: item.news_item_id,
+          total: item.total,
+        )
+        hash
+      end
+  end
+
+  def self.vote_summary(id: nil, total: 0)
```
hanchang: Do we want to allow `nil` id?

hanchang: I presume the `id` refers to a `news_item_id`?

If so, would it be better to have it as an instance method on NewsItem instead?

If not, is the `id` for a `NewsVote`, and if that's the case, would it be better to have it as an instance method on `NewsVote` instead?

app/models/news_vote.rb
```diff
@@ -0,0 +1,104 @@
...
+    {
+      id: id,
+      total: total,
+    }
+  end
+
+  # RETURNS: record if successful, falsey if failed
+  def self.cast_vote(user, news_item_id, vote_direction, multi_vote = false)
```
hanchang: Same question - could this be better as an instance method on `NewsItem` instead of a class method on `NewsVote`?

app/models/news_vote.rb
```diff
@@ -0,0 +1,104 @@
...
+      @result && news_vote
+    rescue ActiveRecord::RecordNotUnique
+      retry
+    rescue ArgumentError => e
+      puts e
+      return
+    end
+  end
+
+  def process_vote(new_vote, multi_vote = false)
```
hanchang: What's a multivote?

app/models/news_vote.rb
```diff
@@ -0,0 +1,104 @@
...
+    end
+
+    if vote == new_vote
+      return 'neutral'
+    else
+      return new_vote
+    end
+  end
+
+  def vote_summary
```
hanchang: This feels like it might be better suited as an instance method on NewsItem instead?

anthotsang:
Closed in favour of https://github.com/coinfi/coinfi/pull/696

PR #699, opened 2019-03-08
Implement sign up modal window

anthotsang:
https://app.asana.com/0/902343675126966/694260327885109
https://app.asana.com/0/902343675126966/1113273052076794


.env.example
```diff
@@ -84,4 +84,8 @@ TWITTER_APP_ID=
 TWITTER_CONSUMER_KEY=
 TWITTER_CONSUMER_SECRET=
 TWITTER_SCAN_KEYWORDS=
+GOOGLE_CLIENT_ID=
```
hanchang: Let's keep this file alphabetized please!

app.json
```diff
@@ -74,6 +74,10 @@
     "TWITTER_API_KEY": { "required": true },
     "TWITTER_API_SECRET": { "required": true },
     "TWITTER_APP_ID": { "required": true },
+    "GOOGLE_CLIENT_ID": { "required": true },
```
hanchang: Also alphabetize please

app/views/layouts/_modal_script.html.erb
```diff
@@ -0,0 +1,217 @@
+<script src="https://unpkg.com/micromodal/dist/micromodal.min.js"></script>
+<script>
```
hanchang: Should this even be an ERB file? Might be better to just have this be a .js file that then gets loaded via <script> tags in the layout directly?

I ask because there's a bunch of syntax irregularities in this file that don't get picked up by `prettier`.

config/environment.rb
```diff
@@ -4,12 +4,14 @@
...
-ActionMailer::Base.smtp_settings = {
-  :port           => 587,
-  :address        => 'smtp.webfaction.com',
-  :user_name      => 'coinfi',
-  :password       => ENV.fetch('WEBFACTION_SMTP_PASSWORD'),
-  :domain         => ENV.fetch('ROOT_DOMAIN'),
-  :authentication => :plain,
-}
-ActionMailer::Base.delivery_method = :smtp
+if Rails.env.production?
```
hanchang: Any reason for changing this to production only?

I find it's often good to be able to test emails on a dev environment.

hanchang: Ah I see, using letter_opener instead.

In that case, might be better to just move this to `config/environments/production.rb`, since `environment.rb` should be reserved for configuration that spans all environments.

hanchang: btw, sometimes we want to test whether an email will go to spam and being able to send email on dev helps with that a lot, so you may want to reconsider letter_opener.

config/locales/en.yml
```diff
@@ -34,6 +34,10 @@ en:
     actions:
       new: New
       show: ''
+  activerecord:
+    errors:
+      email:
+        taken: ASDF
```
hanchang: Do we want this?

app/views/layouts/_modal_script.html.erb
anthotsang: No, it was just taking after tracking scripts, but I guess that file was purposely trying to avoid being picked up by prettier.

PR #738, opened 2019-04-12
Fix price data cache timing

anthotsang:
https://app.asana.com/0/902343675126966/1118359772902582

### DIFF
--- a/app/models/coin.rb
+++ b/app/models/coin.rb
@@ -254,7 +254,7 @@ class Coin < ApplicationRecord
   end
 
   def prices_data
-    Rails.cache.fetch("coins/#{id}/prices", expires_in: seconds_to_next_day) do
+    Rails.cache.fetch("coins/#{id}/prices", expires_in: seconds_to_next_day + 1800) do
       url = "#{ENV.fetch('COINFI_POSTGREST_URL')}/cmc_daily_ohcl_prices?coin_key=eq.#{coin_key}&to_currency=eq.USD&order=time.asc"
       response = HTTParty.get(url)
       JSON.parse(response.body)
@@ -262,7 +262,7 @@ class Coin < ApplicationRecord
   end
 
   def sparkline
-    Rails.cache.fetch("coins/#{id}/sparkline", expires_in: seconds_to_next_day) do
+    Rails.cache.fetch("coins/#{id}/sparkline", expires_in: seconds_to_next_day + 1800) do
       url = "#{ENV.fetch('COINFI_POSTGREST_URL')}/cmc_daily_ohcl_prices?coin_key=eq.#{coin_key}&select=close&to_currency=eq.USD&limit=7&order=time.desc"
       response = HTTParty.get(url)
       results = JSON.parse(response.body)
--- a/app/models/coin.rb
+++ b/app/models/coin.rb
@@ -254,6 +254,7 @@ class Coin < ApplicationRecord
   end
 
   def prices_data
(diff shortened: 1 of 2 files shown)


app/models/coin.rb
```diff
@@ -254,15 +254,15 @@ def hourly_prices_data
   end
 
   def prices_data
-    Rails.cache.fetch("coins/#{id}/prices", expires_in: seconds_to_next_day) do
+    Rails.cache.fetch("coins/#{id}/prices", expires_in: seconds_to_next_day + 1800) do
```
hanchang: What's the logic for the 1800 magic number? Would be helpful to get a comment explaining the situation.

Might also be helpful to create a constant and use it for the `sparkline` call as well if the logic is the same.

app/services/check_cmc_ohclv_service.rb
```diff
@@ -153,7 +183,7 @@ def create_view!
         password=#{@etl_db_pass}',
         'SELECT coin_key, to_currency, time, volume_to
           FROM staging.#{@table[:name]}
-          WHERE time >= NOW() - #{@table[:interval]}::INTERVAL'
+          WHERE time >= NOW() - ''#{@table[:interval]}''::INTERVAL'
```
hanchang: Hm what's this syntax? This is new to me!

app/services/check_cmc_ohclv_service.rb
```diff
@@ -101,16 +103,38 @@ def check_coins
...
+        end
+
+        if latest_cached_price_data.blank?
+          @failed_cached_coins << coin_key
+          next
+        end
+
+        latest_cached_timestamp = DateTime.parse(latest_cached_price_data["time"])
+        db_timestamp = DateTime.parse(row["time"])
+        if latest_cached_timestamp != db_timestamp then
```
hanchang: Do we need to parse the time strings prior to comparison? Legit question, I'm not sure, just checking.

app/services/check_cmc_ohclv_service.rb
anthotsang: The double single quotes escape the single quotes because the interval clause is nested inside a single quoted string.
The ::INTERVAL just casts to interval instead of using `interval ''#{@table[:interval]}''`.
This is just a refactor, where the double single quotes were in each table hash before.

app/services/check_cmc_ohclv_service.rb
anthotsang: The cached version is formatted slightly different so a direct string comparison was failing.

app/models/coin.rb
anthotsang: I thought about it, but that constant by itself wouldn't make so much sense I think? I've added a comment that hopefully helps explain it.
I may adjust the value in the future if it proves to be insufficient.

app/services/check_cmc_ohclv_service.rb
```diff
@@ -71,23 +72,24 @@ def call
...
 
       if coins.length == 0
         @failed_tests << coin_test[:title]
         next
       end
 
       coins.each do |coin|
-        coin_key = coin[0].to_s
-        ranking = coin[1]
+        coin_key = coin.coin_key.to_s
```
hanchang: Just curious, when is a `coin_key` not a string?

anthotsang: I honestly can't remember why I had that there. It may have just been because I was performing string operations and wanted to make sure it was a string? It can probably be removed but I didn't want to break anything haha.

PR #752, opened 2019-04-26
Github stats for indicators

anthotsang:
https://app.asana.com/0/902343675126966/850764393402017


app/helpers/indicators_helper.rb
```diff
@@ -26,4 +26,8 @@ def get_buy_sell_class(value)
       'neutral'
     end
   end
+
+  def github_url(repo)
+    "https://www.github.com/#{repo}"
```
hanchang: I think their default is to not include the `www` subdomain so might be good to remove it here as well.

app/models/coin.rb
```diff
@@ -33,7 +33,7 @@ class Coin < ApplicationRecord
   has_many :calendar_event_coins
   has_many :calendar_events, through: :calendar_event_coins
 
-  validates :name, uniqueness: true, presence: true
+  validates :name, presence: true
```
hanchang: Hm why did the unique validation get removed?

app/models/coin.rb
```diff
@@ -276,6 +276,24 @@ def sparkline
     end
   end
 
+  def github_stats(force_refresh: false)
+    return nil unless github_repo.present?
```
hanchang: I think you can just do `return` instead of `return nil` to accomplish the same thing.

app/models/coin.rb
```diff
@@ -276,6 +276,24 @@ def sparkline
...
+
+    Rails.cache.fetch("coins/#{id}/github_stats", force: force_refresh) do
+      response = CoinServices::RetrieveGithubStats.call(coin: self)
+      response.try(:result)
+    end
+  end
+
+  def merge_github_stats(stats_to_merge)
+    return nil unless github_repo.present?
```
hanchang: I think you can just do `return` instead of `return nil` to accomplish the same thing.

app/models/coin.rb
anthotsang: There were already violations in the database and as far as I can tell, there's no requirement for it?

app/services/coin_services/retrieve_github_stats.rb
```diff
@@ -0,0 +1,77 @@
...
+        code_frequency: code_frequency,
+        snapshot: snapshot,
+      }
+    end
+
+    private
+
+    def retrieve_commit_activity_data(repo_path)
+      data = @client.commit_activity_stats repo_path
+      return nil unless data.present?
```
hanchang: Still some `return nil` statements but whatevs

app/services/coin_services/retrieve_github_stats.rb
```diff
@@ -0,0 +1,77 @@
...
+        {
+          timestamp: timestamp,
+          commits: commits,
+        }
+      end
+    end
+
+    def retrieve_code_frequency_data(repo_path)
+      data = @client.code_frequency_stats repo_path
+      return nil unless data.present?
```
hanchang: Still some `return nil` statements but whatevs

app/services/coin_services/retrieve_github_stats.rb
```diff
@@ -0,0 +1,77 @@
...
+          timestamp: weekly_data[0],
+          additions: weekly_data[1],
+          deletions: weekly_data[2],
+        }
+      end
+    end
+
+    def retrieve_repository_stats(repo_path)
+      repository = @client.repo repo_path
+      return nil unless repository.present?
```
hanchang: Still some `return nil` statements but whatevs

PR #779, opened 2019-06-21
Add last modified to indicators

anthotsang:
https://app.asana.com/0/902343675126966/1128043521235536


app/models/coin.rb
```diff
@@ -251,6 +251,7 @@ def token_metrics
 
   def hourly_prices_data
     Rails.cache.fetch("coins/#{id}/hourly_prices", expires_in: seconds_to_next_hour) do
+      self.touch
```
hanchang: Not sure tbh but could it be better to do the `touch` at the end in case anything fails and also for performance?

I notice you do it for some methods but not others so there's probably a reason that I'm just missing...

PR #784, opened 2019-06-27
Add indicator api

anthotsang:
https://app.asana.com/0/902343675126966/1128634317955899


config/routes.rb
```diff
@@ -41,7 +41,7 @@
     resources :contributor_submissions, path: 'contributor-submissions'
     resources :exchange_listings, only: :index, path: 'listings'
     resources :calendar_events, only: :index, path: 'calendar'
-    get '/indicators/:id_or_slug', to: 'indicators#show', as: 'indicators'
+    get '/indicators/:symbol', to: 'indicators#show', as: 'indicators'
```
hanchang: Does order matter? Might be better to put this line after the `/indicators/overview/` entry to ensure that the hardcoded URLs get evaluated first.

anthotsang: It's not at all obvious from the way github has presented it, but this one is just `/indicators` whereas the other two are actually `/api/indicators/*`.

hanchang: 🤦‍♂ Rookie mistake on my part, thanks for the heads up!

PR #794, opened 2019-07-04
Change coin stats for indicators

anthotsang:
https://app.asana.com/0/902343675126966/1129841088787595


app/controllers/indicators_controller.rb
```diff
@@ -72,6 +73,10 @@ def set_indicator_data
     @summary_value = calculations[:summary_value]
   end
 
+  def set_coin_stats
+    @coin_stats = Rails.cache.read("indicators/#{@coin.slug}:stats")
```
hanchang: Dumb question - what happens if the cache isn't populated?

Would recommend lazy loading the `CoinService` call here if the cache is empty just in case, that's generally a best practice when it comes to caching logic - the only exception is if the job takes entirely too long for an HTTP request to deal with, in which case there should be reasonable default value returned instead.

anthotsang: It would just omit the section. Will add the call to coin service.

app/controllers/indicators_controller.rb
```diff
@@ -72,6 +73,16 @@ def set_indicator_data
...
```
hanchang: Sorry for nitpicking but the best practice is to use `Rails.cache.fetch` and pass in the block that represents what happens upon cache miss.

https://api.rubyonrails.org/classes/ActiveSupport/Cache/Store.html#method-i-fetch

hanchang:
This is good to merge after the transition to using `Rails.cache.fetch`, nicely done!

app/controllers/indicators_controller.rb
hanchang: Just to be explicit, this way of operating is important because it lazily populates the cache for subsequent requests, whereas in your current implementation it is possible for the cache to never be populated nor read.

app/controllers/indicators_controller.rb
```diff
@@ -72,6 +73,18 @@ def set_indicator_data
     @summary_value = calculations[:summary_value]
   end
 
+  def set_coin_stats
+    @coin_stats = Rails.cache.fetch("indicators/#{@coin.slug}:stats") do
+      fetched_stats = CoinServices::UpdateIndicatorStats.call.result
+
+      if fetched_stats[@coin.slug].present?
```
hanchang: I *think* you can further simplify this entire `if/else` block to just

```
    fetched_stats.fetch(@coin.slug) || nil
```

Might be good to protect against the case where the coin slug doesn't exist in `fetched_stats` for whatever reason.

hanchang: But that said maybe we do want to throw an exception for Rollbar to pick up to alert us about the problem...

PR #883, opened 2020-01-30
Amp pages

anthotsang:
https://app.asana.com/0/813012001853927/1159492027039817


app/controllers/coin_articles_controller.rb
```diff
@@ -1,14 +1,15 @@
...
   rescue_from ActiveRecord::RecordNotFound, :with => :record_not_found
   before_action :set_coin_article, only: [:show]
+  prepend_before_action :set_amp, only: [:index]
 
   breadcrumb 'How to Buy Cryptocurrency', :coin_articles_path, match: :exact
 
   def index
     distribute_reads(max_lag: MAX_ACCEPTABLE_REPLICATION_LAG, lag_failover: true) do
       @coin_articles = CoinArticle.page(params[:page]).per(params[:limit])
+      @coin_articles.each # force load
```
hanchang: Thanks for the comment! Just curious why force loading is required here?

app/views/components/_rollbar.html.erb
```diff
@@ -0,0 +1,15 @@
+<script>
```
hanchang: I'm guessing this is for Rollbar client side JS?

config/initializers/rollbar.rb
```diff
@@ -80,16 +80,5 @@
...
-    accessToken: ENV.fetch('ROLLBAR_CLIENT_ACCESS_TOKEN'),
-    captureUncaught: true,
-    captureUnhandledRejections: true,
-    logLevel: 'error',
-    hostWhiteList: ['coinfi.com', 'herokuapp.com'],
-    payload: {
-      environment: "production"
-    }
-  }
+  config.js_enabled = false
```
hanchang: Ah interesting, so the Rollbar `js_enabled` functionality doesn't work out of the box and requires custom integration?

app/controllers/coin_articles_controller.rb
anthotsang: distribute_reads needs to force load to work properly. normally the recommended method is `to_a`, but that would get rid of the methods provided by Kaminari
distribute_reads also suggests `load`, but that didn't seem to be working either

app/views/components/_rollbar.html.erb
anthotsang: Yep

config/initializers/rollbar.rb
anthotsang: No, it works, but the problem was it was enabled for everything and couldn't be selectively disabled.
So, it was being injected into AMP pages, which can't support external javascript.

anthotsang:
> Also, I'm guessing we're only applying AMP to the coin_articles index page at the moment because we're rolling it out slowly to test before doing some sort of sitewide (or at least coin_articles-wide) release?

Yep! Testing bit-by-bit to hopefully get a better idea of what the effects are.

PR #895, opened 2020-02-21
Exchange Reviews

anthotsang:
https://app.asana.com/0/813012001853927/1162193408125705


app/controllers/application_controller.rb
```diff
@@ -31,6 +31,10 @@ def set_amp_canonical
 
 protected
 
+  def record_not_found
```
hanchang: Nice refactor 💪

